PR #460, opened 2020-09-21
fix: add missing county level to quidel

### LINKED ISSUE #539: Enable COVID antigen test signals at the county level
@jingjtang says the COVID antigen test signal is now available at 700 or more counties, meaning it could be useful to allow it to be mapped.

Currently the map is inconsistent. We do not allow the user to select the antigen tests to map at the county level, but since county data is in the API,  viewing the county data for another signal will show the antigen tests in the list of graphs on the right. So you can view county data easily, you just can't see a map of it.

I propose including county maps in the next release.

sgratzl:
according to meta data we have it

closes #539 

**Prerequisites**:

- [x] Build is successful

### Summary

according to meta data we have it

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "delphi-hcii-covidcast",
-  "version": "1.8.2",
+  "version": "1.9.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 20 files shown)


tildechris (approved):
Since we now have #544 to explicitly mention coverage, I think it's reasonable to enable this signal at the county level.

PR #497, opened 2020-09-24
Style tweak bug fixes

ejfox:
style(bug): fixes #487
style(bug): fixes #476
style(bug): fixes #475
style(bug): fixes #473
style(bug): fixes #492 by setting max-width for legend and options
style: fixes #494
style: fixes #480 and increases default locations, shows date

Screen Shot 2020-09-24 at 1 03 02 PM
Screen Shot 2020-09-24 at 1 03 18 PM
Screen Shot 2020-09-24 at 1 03 34 PM

Issue #488, opened 2020-09-24
improve website menus levels


sgratzl:
from @jennyxxdance:
> I wanted to bring this up because I noticed there seems to have three different rows of tabs and I’m wondering how we could bring the UI a little better

<img width="1418" alt="Screen Shot 2020-09-23 at 9 26 05 AM" src="https://user-images.githubusercontent.com/4129778/94135293-8e400880-fe63-11ea-9553-a2edfc45966e.png">

sgratzl:
there are different alternatives already in some PRs:

 https://github.com/cmu-delphi/www-covidcast/pull/506

 ![image](https://user-images.githubusercontent.com/4129778/94423158-7aa8e080-0188-11eb-916d-04ce44a2a72b.png)

https://github.com/cmu-delphi/www-covidcast/pull/445

![image](https://user-images.githubusercontent.com/4129778/93307708-96fc6300-f801-11ea-93c0-63057be808bf.png)

sgratzl:
closing in favor of the new website design

PR #506, opened 2020-09-28
change menu header style

sgratzl:
extracted from #497 

image

### DIFF
--- a/src/components/ModeNav.svelte
+++ b/src/components/ModeNav.svelte
@@ -5,7 +5,27 @@
 
 <style>
   .mode-container {
+    margin: 0;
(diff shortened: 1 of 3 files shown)


tildechris:
Actually, I think the tab style gives some affordance that it can be clicked, and it's not super visually clear that the black text changes the map and the red text is navigation.  Let's take this to UX review.

sgratzl:
this PR is proposal by EJ for the issue: https://github.com/cmu-delphi/www-covidcast/issues/488 in which the main discussion is happening

sgratzl:
closing for, waiting for redesign

PR #526, opened 2020-09-30
feat: add background to map title for visibility

### LINKED ISSUE #512: Map titles hard to read against map background
**Actual Behavior:**

When I select a county and the map zooms me in to that county, the map title can be hard to read against the background, depending on the map colors behind it.

1. go to https://cms-staging.andrew.cmu.edu/covid/?sensor=fb-survey-smoothed_hh_cmnty_cli&level=county&date=20200925&signalType=value&encoding=color&mode=overview&region=49049
2. look at the map title

**Expected behavior**

Some way to make it visible against the background. I'm not sure what would work; putting the title in another bubble/block may feel cluttered.

Also, it appears the title is forced to be 12px font, while the other navigation links on the top of the page are 16px, and the graph titles on the right are 1em/14px, so the map title is nearly the smallest text on the entire page.

**Screenshots**
Screen Shot 2020-09-29 at 2 27 20 PM

3

**Context**

- Browser: Firefox 81

kateharwood:
closes #512 

**Prerequisites**:

- [x] Build is successful

### Summary
I created this pull request mainly to remind myself of how the git process works and get familiar with the codebase. If this is not the way you would like to fix this issue totally fine to just reject this.

Adding background class to map title.
Tweaking desktop @media rule so that the map title has a background, rounded on mobile, flat no edges on desktop. 

DESKTOP
Before
Screen Shot 2020-09-30 at 10 26 15 AM

After
Screen Shot 2020-09-30 at 10 57 34 AM

MOBILE
Before
Screen Shot 2020-09-30 at 11 00 13 AM

After
Screen Shot 2020-09-30 at 11 16 55 AM

### DIFF
--- a/src/components/MapOverlays.svelte
+++ b/src/components/MapOverlays.svelte
@@ -46,7 +46,6 @@
   /** desktop **/
   @media only screen and (min-width: 767px) {
     .title-container {
-      background-color: unset;
       box-shadow: none;
     }
   }
--- a/src/components/MapOverlays.svelte
+++ b/src/components/MapOverlays.svelte
@@ -38,7 +38,7 @@
     display: flex;
     align-items: flex-start;
     justify-content: center;
-    margin-top: 2px;
+    margin: 2px 5px;
   }
   .signal-description {
     margin-bottom: 0.25em;
--- a/src/components/MapOverlays.svelte
+++ b/src/components/MapOverlays.svelte
@@ -49,8 +49,6 @@
     .title-container {
       box-shadow: none;
       border: none;
-      border-radius: 0;
-      margin-top: 0;
     }
   }
 
--- a/src/components/MapOverlays.svelte
+++ b/src/components/MapOverlays.svelte
@@ -38,6 +38,7 @@
     display: flex;
     align-items: flex-start;
     justify-content: center;
+    margin-top: 2px;
   }
   .signal-description {
     margin-bottom: 0.25em;
@@ -47,6 +48,9 @@
   @media only screen and (min-width: 767px) {
     .title-container {
       box-shadow: none;
+      border: none;
+      border-radius: 0;
+      margin-top: 0;
     }
   }
 
--- a/src/components/MapOverlays.svelte
+++ b/src/components/MapOverlays.svelte
@@ -68,7 +68,7 @@
 </style>
 
 <div class="map-top-overlay">
-  <div class="title-container">
+  <div class="title-container container-bg">
     <Title {showDate} />
   </div>
   <div class="map-controls-container">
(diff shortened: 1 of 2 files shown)


sgratzl (commented):
what is the reason you chose to use rounded borders in mobile but not in the desktop version?

Moreover, the issue also talks about increasing/matching the font-size but I cannot see that in this PR

kateharwood:
Hey Sam,
1. Rounded borders on mobile but not desktop: I saw that the rounded border on mobile used to be there in previous versions, but thought that it looked better without the border and rounded corners on desktop (it seemed to me more subtle and less in the way when zooming in and viewing the map). Totally happy to change that if you think it's better for mobile and desktop to be the same.

2. Text sizing: I took the mention of text sizing in the issue as one of the reasons that the title was not easily readable, and not an issue itself. Totally happy to change that as well though if the text sizing itself is an issue.

sgratzl:
I can just comment on the code but I don't know the desired behavior, adding @adamperer for his input

adamperer:
Thanks @kateharwood for jumping in and making a contribution!  Right now, some of these design decisions are admittedly being made ad-hoc -- hopefully Chris can help us coming up with a better process to support the growing team -- so it doesn't rely solely on my (arbitrary) judgment 😄 

I definitely think your contribution is an improvement over the current status of the text being challenging to read.    It may make sense to use rounded corners in both designs, as the straight edges seem a bit rough and choppy in Firefox:
image

Keeping the text small may make sense in this case, to avoid cutting off too much of the map.  @capnrefsmmat , do you think this latest design fixes your issue #512 ?

Again, thanks @kateharwood for taking this initiative!

capnrefsmmat:
Hmm, yes, I think this is definitely an improvement.

Regarding rounded corners, I think it could go either way:

* If the white box spanned full width, so you didn't see bits of map peeking around the corners, then square corners are fine, because there are no corners.
* But if the box is going to span only most of the width, I'd have it be rounded corners. In fact, maybe it can merge into the box above, instead of being a separate rounded thing. (Maybe; merging would also make it feel less like a map title and more like a description for those drop-downs)

Regarding font size, I have no strong opinions. I suspect the right person to ask about 12px font is someone who needs reading glasses. The *combination* of poor contrast with background and small size definitely made it hard to read, but now that the background is white, it may be better.

kateharwood:
Good points! For now I updated desktop to have rounded corners as well. I also tweaked the margin on mobile of the search box so it doesn't butt up against the map.

Seems like merging the title into the box above on desktop also could make sense. 

Updated after pics:
Screen Shot 2020-09-30 at 2 23 27 PM

Screen Shot 2020-09-30 at 2 27 56 PM

tildechris (approved):
LGTM.  The only feedback that I have is perhaps the rounded-box doesn't need to span the entire width of the map, but I don't have a strong intuition if that's better in all cases.

PR #530, opened 2020-10-01
feat: generate signal constants

### LINKED ISSUE #522: Improved method to automatically import strings document into code
It would be great if there could be an automated way to import all of the text of the indicators.  Previously, [this document](https://docs.google.com/document/d/13naRh47E6OjiHwV2z-VyU7o9vE6MlSj_cF6x2u9pdbc/edit) was copy and pasted directly into the code and was a less painful way to import the text.

However, some of the text migrated [here](https://docs.google.com/document/d/1kDqRg8EaI4WQXMaUUbbCGPlsUqEql8kgXCNt6AvMA9I/edit#) and this process broke.

@sgratzl, could you suggest a better way to allow collaborate edits outside the code (with similar ease to Google Docs), but make it less painful to import in the code base.

Some possibilities may include google sheets (where you can download a CSV), a structured markdown document, or go back to the JSON format we had previous?  Open to any ideas you may have to make your life easier.

Tagging @RoniRos and @capnrefsmmat to follow the discussion.

sgratzl:
closes #522

**Prerequisites**:

- [x] Build is successful

### Summary

implements a custom logic at build time to load and parse: https://docs.google.com/document/d/1RLy4O-gtACVjLEVD_vxyPqvp9nWVhqEjoWAKi68RKpg/edit# which is a mix of YAML/Markdown document for generating the signal list and description. 

Notes
 * this run on build time. So, when the app is deployed or locally called (`npm run gen`)
 * the document reflects the current constants.js state with the latest document descriptions
 * it will always use the latest version of this document in plain TXT form. There is no such thing as tagging or branching.
 * the document has to be publicly viewable (anyone with the link can view the document)


krivard:
Intriguing! Question on the precise definition of "latest version": If we make edits to the linked document with track changes enabled, will that (1) crash, (2) pull as if all changes were rejected, (3) pull as if all changes were accepted?

sgratzl:
as mentioned in the google doc itself, the script is just downloading the file as a TXT file. I think adding more revision logic requires using the proper google api which itself requires API access and register the script as an proper app.

tildechris:
I think this is a neat approach, but I worry about the brittleness.  How often does this content change?  Would it not be more appropriate to this markdown be in a text file and not in docs?  We'd then be able to control it as a project asset directly.

I don't think we have to address these question on this PR since this looks like an improvement over the status quo, but I'll add an issue to investigate what a structured solution would look like.

tildechris (approved):
LGTM.

tildechris (changes requested):
@sgratzl as discussed on #522/#570 can you save the literal text content of the doc that's pulled down into a separate build asset that gets included in each release?

sgratzl:
> @sgratzl as discussed on #522/#570 can you save the literal text content of the doc that's pulled down into a separate build asset that gets included in each release?

should it be a built artifact so just part of the `release.zip` or really commited to the repository. In the later case, who should commit it when it changes?. Or do we just wanna run the generation on demand and do it manually?. Should I store the raw text or the generated JSON file?

tildechris:
If there's a good way to get it actually committed to the repo, I think we should consider it, but considering when to check that in made me feel like that process would be error prone.  Let's defer on that discussion for now so that we can close this PR out.

> Should I store the raw text or the generated JSON file?

I think we should store the raw text in the release.zip.

sgratzl:
ok, I changed it that files like this one will be generated and also copied to output directory = part of the release.zip:

[descriptions.raw.txt](https://github.com/cmu-delphi/www-covidcast/files/5402980/descriptions.raw.txt)

tildechris (approved):
Thanks Sam!  I tested this locally and replicated the intended behavior.  I notice that the build time is longer, so we'll want to keep an eye on that in the future.

LGTM.

PR #543, opened 2020-10-07
feat: enforce county geo type

### LINKED ISSUE #482: define which geo level to use for scales
for e.g. doctors visits only one scale (mean, stddev) independent of the geo type (state, county) is used. 

However, taking a look at the code it is not defined which geo type that is, it seems to use the first one coming back from the api, which is unreliable:

https://github.com/cmu-delphi/www-covidcast/blob/ef598a313df7b632b795c202c5d55f3f52d987ad/src/data/index.js#L66-L82

possible options:
 * use counties as reference
 * use the maximal one (defined by mean)?
 * use an average?

sgratzl:
for computing regular signal meta data

closes #482

**Prerequisites**:

- [x] Build is successful

### Summary

there is no visual change but it makes the front end independent of the meta data order returned by the API


src/data/index.js
```diff
@@ -65,9 +72,18 @@ function updateStatsMap(key, matchedMeta, statsMap) {
...
 function loadRegularSignal(sEntry, meta, timeMap, statsMap) {
-  const matchedMeta = meta.epidata.find(
-    (d) => d.data_source === sEntry.id && d.signal === sEntry.signal && (!d.time_type || d.time_type === 'day'),
-  );
+  // find the matching meta data by looping through the candidates and fallback to the first one
+  const baseFilter = (d) =>
+    d.data_source === sEntry.id && d.signal === sEntry.signal && (!d.time_type || d.time_type === 'day');
+  const candidates = regularSignalMetaDataGeoTypeCandidates.slice();
+  let matchedMeta = null;
+  while (!matchedMeta && candidates.length > 0) {
```
tildechris, line 80: Nit: (Optional to fix) this is a bit hard to read, consider a single expression, something like:

`meta.epidata.filter((d)=>baseFilter(d)).sort(sortByArrayOrder)[0]` where sortByArrayOrder chooses "msa" then "county" otherwise "state" in that order, or choose a data structure that has an explicit order between levels.

tildechris (approved):
LGTM.

src/data/index.js
sgratzl: something like in the latest update?

src/data/index.js
```diff
@@ -75,15 +75,26 @@ function loadRegularSignal(sEntry, meta, timeMap, statsMap) {
...
+      // missing is bigger
+      return 1;
+    }
+    if (bIndex < 0) {
+      return -1;
+    }
+    return aIndex - bIndex;
+  };
+
+  const candidates = meta.epidata.filter(baseFilter).sort(byGeoTypePriority);
```
tildechris, line 96: Yes! Thanks, I like this a lot.

PR #544, opened 2020-10-07
coverage info

### LINKED ISSUE #538: consider adding a summary stats view
It would be great if a summary statistics would be shown somewhere close to the map for the current view. 

statistics might include
 * number/percentage of counties where data is valid i.e. not missing
 * number/percentage of counties where the value is 0
 * since we use mega counties a lot: how many states are augmented with mega-county information
 * min/max/median/average value?

sgratzl:
based on #540 for simpler development

closes #538 

**Prerequisites**:

- [x] Build is successful

### Summary

adds a small field to the bottom right of the map showing the "coverage":

image

a tooltip explains it:

image

coverage is the percentage of valid level (e.g., county) data for the current signal and date. For example in the given screenshot it shows that the map has only 'Doctor Visits' data for 61% of all counties. Mega Counties are explicitly excluded. 

In the issue I suggested more value but I think the coverage is the most important one

### DIFF
--- a/src/components/MapOverlays.svelte
+++ b/src/components/MapOverlays.svelte
@@ -7,12 +7,14 @@
   import ColorLegend from './legends/ColorLegend.svelte';
   import BubbleLegend from './legends/BubbleLegend.svelte';
   import SpikeLegend from './legends/SpikeLegend.svelte';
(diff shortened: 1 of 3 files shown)


tildechris (approved):
LGTM.

PR #546, opened 2020-10-07
single location view

sgratzl:
based on #540 for simpler development

**Prerequisites**:

- [x] Build is successful

### Summary

adds a new 'single location' mode/tab which show data similar to the side panel:

image

In addition, when hovering also the key facts of all cards are updated to show their current value:

image

In addition, it support a simple comparison: 

image


tildechris:
My first reaction is that showing a number of time-series is a bit too low level.  Compare with CAN, where there is a single data point per signal along with a color-coded indication of "goodness".  We could show present value, how much it is moving (7-day percentage diff), and perhaps a sparkline showing historical movement.

For example:

Cases|Deaths
--|--
15 (⇧10% 7-day) `__==__==**` | 2 (⇧20% 7-day) `***==--__**`

Away From Home|Doctors Visits|Covid-Like Symptoms
--|--|--
0.04 (⇧10% 7-day) `__==__==**` | 5.6% (⇧20% 7-day) `***==--__**` | 0.7% (⇧20% 7-day) `_____==++##**`

I think there's value in the correlated time-series charts since people will attempt to use them to see correlations, but maybe that should be a detail view?

philmcguinn:
I don't have any issues with the current approach. I do have a question though - is the comparison section hover-able as well? My natural inclination is to try to compare two regions at a past point in time when I see those graphs, so it could be good to allow that if it's not too much work.

Letsgohyde:
Hey guys, without context it's hard of me to leave feedback, but if it's as simple as adding a tab, LGTM. 

fyi: For features moving forward, I suggest we list the user goals or needs. Also any docs, requirements that can help add context :)

tildechris (approved):
Let's go ahead with this approach as-is for now, and we'll separately start a single-location view discussion.

PR #454, opened 2020-09-18
create staging environment setup

sgratzl:
**Prerequisites**:

- [x] Unless it is a hotfix it should be merged against the `dev` branch
- [x] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [x] Build is successful
- [x] Code is cleaned up and formatted

### Summary

allows to specify the covidcast endpoint api that should be used during **build time** using an environment variable named `COVIDCAST_ENDPOINT_URL`.  e.g., when setting the env variable on netlfiy, see `netlify.toml` file to use a potential staging server for dev and deployment previews

PR #547, opened 2020-10-07
POC for configuring the preview environment to use our staging endpoint proxy

korlaxxalrok:
**Prerequisites**:

- [X] Build is successful
- [X] Code is cleaned up and formatted

### Summary

(I'm not very well plugged into the dev workflow for this repo, so apologies in advance for this verbose PR.)

This can be considered a proof-of-concept PR, so if you eventually want to scrap it and take the useful parts and use them in another PR I won't mind :) My additions were partly cribbed from [this PR](https://github.com/cmu-delphi/www-covidcast/pull/454), so it may be easier/cleaner to update that one, or someone may want to validate this locally first, etc.

The goal is to be able to more easily view prototype signals in the preview application. To do this we need to add some configuration that will send the Netlify API requests to a staging endpoint proxy that we have recently created. This proxy has the ability to route requests a little more flexibly, e.g. route a prototype signal request to our staging environment, otherwise route to production.

To test locally, I first created a fake signal on our staging server, added it to `src/stores/constants.js`, and then updated `const ENDPOINT = 'https://api-staging.delphi.cmu.edu/epidata/api.php';` in `src/data/api.js` (wasn't sure how to make the local dev server use this staging endpoint proxy without this update and I haven't actually included that change in the PR).

**The result in local dev is that the test signal (which is only available in our staging environment) is shown as expected:**

image

Additionally, all other signals were visible, which indicates that the other non-prototype (non-staging) signals were being loaded from the production endpoint.

I think our ask is that this PR, or at least the good parts from it, can be merged so that we can deploy the preview application to further validate that this scheme generally works.


sgratzl:
without https://github.com/cmu-delphi/www-covidcast/pull/454 I don't see that this PR in its current state would work. Moreover, I'm wondering what the strategy is which branches/PR will be pointing at the staging one and which are not. I wanna avoid that we keep developing against the staging one and as soon as we publish (merge to main) notices that several signals doesn't work anymore. 

Moreover, what is the plan for this regarding https://github.com/cmu-delphi/www-covidcast/issues/522 and https://github.com/cmu-delphi/www-covidcast/pull/530 ?

krivard:
> without #454 I don't see that this PR in its current state would work.

That's probably just a flubbed workflow -- @korlaxxalrok would you either change the base branch to sgratzl/configure_endpoint, or rebase onto sgratzl/configure_endpoint so that those changes are included in this PR?

> Moreover, I'm wondering what the strategy is which branches/PR will be pointing at the staging one and which are not. 

I may not be clear on what you're asking -- were you not intending in #454 that (1) all branches would have the same 3-flavor netlify endpoint configuration, and that (2) the 3 flavors (production, branch-deploy, deploy-preview) could represent three distinct endpoint addresses?

> I wanna avoid that we keep developing against the staging one and as soon as we publish (merge to main) notices that several signals doesn't work anymore.

The proposed workflow is this:

There are three servers: the production API, the staging API, and a proxy. The proxy dispatches queries to production or staging based on source and signal name regexes.

1. Engineering develops a new signal prototype.
2. Engineering uploads the signal prototype to the staging API.
3. Engineering adds the signal prototype regex to the proxy, so that queries for the new signal will be routed to the correct location.
4. Engineering opens a PR adding the prototype to constants.js
5. Engineering circulates the visualization of the prototype for review via the 'Details' link in the PR
6. Engineering iteratively revises the signal prototype based on feedback
7. When signal prototype is approved, **Engineering promotes it to the public API server, and removes the regex from the proxy**
8. Final review of PR via the 'Details' link; Engineering verifies that data is being retrieved from public
9. Viz merges PR

> Moreover, what is the plan for this regarding #522 and #530 ?

That is a very good question, and we should discuss. This PR is incompatible with removing the signal configuration from constants.js.

korlaxxalrok:
@krivard Updated the base branch.

sgratzl:
one possible way:
 * we add the possibility that the google doc PR not just reads from one google docs but multiples ones and they are concatenated
 * we use a special branch (similar to the dev branch), having its own deploy context (https://docs.netlify.com/site-deploys/overview/#deploy-contexts) which specify that it should use your staging server and a secondary google doc. 

Thus, if you want to test a new signal: 
 * you edit the extra google doc 
 * trigger a redeploy of special branch see https://docs.netlify.com/configure-builds/build-hooks/

sgratzl:
I start adding support for specifying (multiple) google docs in https://github.com/cmu-delphi/www-covidcast/pull/530. The environment variable is: `COVIDCAST_SIGNAL_DOC` which can be a comma separated list of public google doc links

sgratzl:
closing in favor of the extended #454

PR #555, opened 2020-10-09
feat: add mask signal

### LINKED ISSUE #554: Add smoothed_wearing_mask signal
In order to coordinate some communications going out on Monday, we would like to add the `smoothed_wearing_mask` signal as a point release 1.9.3.

sgratzl:
closes #554

**Prerequisites**:

- [x] Build is successful

### Summary

image

open questions:
 * atm. the lowest color is 66% and not 0. The reason is as noted in another issue that it is not enforced in the choropleth map that 0 = first value

notes:
 * since it wasn't specified I added it into the 'Public's Behavior' category

image

### DIFF
--- a/src/stores/constants.js
+++ b/src/stores/constants.js
@@ -255,6 +255,28 @@ const defaultSensors = [
   },
   {
     type: 'public',
+    name: 'People Wearing Masks',
+    longDescription:
+      // prettier-ignore
+      `Percentage of people who wear a mask most or all of the time while in public, based on surveys of Facebook users.`,
+    links: [
+      {
+        href: 'https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/fb-survey.html',
+        alt: 'Technical description',
+      },
+    ],
+    id: 'fb-survey',
+    signal: 'smoothed_wearing_mask',
+    levels: ['county', 'msa', 'state', 'hrr'],
+    mapTitleText:
+      // prettier-ignore
+      'Percentage of people who wear a mask most or all of the time while in public, based on surveys of Facebook users',
+    yAxis: 'Percentage',
+    format: 'percent',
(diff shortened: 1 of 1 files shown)


src/stores/constants.js
```diff
@@ -253,6 +253,28 @@ const defaultSensors = [
     format: 'raw',
     hasStdErr: true,
   },
+  {
+    type: 'public',
+    name: 'People Wearing Masks',
+    longDescription:
+      // prettier-ignore
+      `Percentage of people who wear a mask most or all of the time while in public, based on surveys of Facebook users.`,
```
capnrefsmmat, line 261: This should be the text from the doc:

> Every day, Delphi surveys tens of thousands of Facebook users, asking them a broad set of COVID-related questions, including whether they, or anyone in their household, are currently experiencing COVID-related symptoms. We also ask them if they wear a mask when they are in public. For this signal, we estimate the percentage of people who wear a mask most or all of the time when they are in public.

capnrefsmmat:
>  atm. the lowest color is 66% and not 0. The reason is as noted in another issue that it is not enforced in the choropleth map that 0 = first value

I think that's actually good here. Since this signal is always > 50% (at least, right now it is), anchoring at 0 would make the map hard to read.

> since it wasn't specified I added it into the 'Public's Behavior' category

I think that's a good placement.

sgratzl:
> I think that's actually good here. Since this signal is always > 50% (at least, right now it is), anchoring at 0 would make the map hard to read.

at least on regular line/bar charts the rule is that you should start the scale at 0. it depends what you wanna show, starting at 0 will show that overall the values are great in the US (here denoted as red) (even tho great = good), starting at 66% will highlight the variance within the US

sgratzl:
updated to interpolateYlGnBu color scale

image

PR #556, opened 2020-10-09
add download button to detail views

### LINKED ISSUE #484: add download options to detail views
It would be great if the vega views (such as the detail view) has a download option in PNG or SVG format. Since both are a standard feature of Vega it should be a simple one. 

image

e.g. a start is in 

https://github.com/cmu-delphi/www-covidcast/blob/ce717edd3ec89a78e56c6a6e7ab4e9ac6149d15a/src/components/DetailView/DetailView.svelte#L149-L156

dlaliberte:
closes #484

**Prerequisites**:

- [x] Build is successful

### Summary

Add button to download image for Details chart.

### DIFF
--- a/src/components/DetailView/DetailView.svelte
+++ b/src/components/DetailView/DetailView.svelte
@@ -10,6 +10,8 @@
   import { trackEvent } from '../../stores/ga';
   import VegaTooltip from './VegaTooltip.svelte';
   import IoMdHelp from 'svelte-icons/io/IoMdHelp.svelte';
+  import IoIosSave from 'svelte-icons/io/IoIosSave.svelte';
+  import { downloadUrl } from '../../data/screenshot';
 
   const dispatch = createEventDispatcher();
   /**
@@ -76,6 +78,18 @@
       close.focus();
     }
   });
+
+  // Reference to the vega chart component.
+  let vegaRef = null;
+
+  function downloadVega() {
+    vegaRef.vegaAccessor().then((view) => {
+      const pngP = view.toImageURL('png', 2);
+      pngP.then((url) => {
+        downloadUrl(url, `${sensor.name} in ${'Unknown'}.png`);
+      });
+    });
+  }
 </script>
 
 <style>
@@ -150,6 +164,13 @@
           currentInfoSensor.set(sensor);
         }}><IoMdHelp /></button>
     {/if}
+    <button
+      title="Download this view"
+      class="pg-button pg-button-circle info"
+      on:click={downloadVega}
+      disabled={!vegaRef}>
+      <IoIosSave />
+    </button>
   </div>
   <button
     bind:this={close}
@@ -164,6 +185,7 @@
 </div>
 <div class="single-sensor-chart vega-wrapper">
   <Vega
+    bind:this={vegaRef}
     {data}
     {spec}
     {patchSpec}
--- a/src/components/DetailView/DetailView.svelte
+++ b/src/components/DetailView/DetailView.svelte
@@ -1,11 +1,5 @@
 <script>
-  import {
-    signalCasesOrDeathOptions,
-    currentDateObject,
-    smallMultipleTimeSpan,
-    currentInfoSensor,
-    currentRegionInfo,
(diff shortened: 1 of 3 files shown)


src/components/DetailView/DetailView.svelte
```diff
@@ -76,6 +78,18 @@
...
   });
+
+  // Reference to the vega chart component.
+  let vegaRef = null;
+
+  function downloadVega() {
+    vegaRef.vegaAccessor().then((view) => {
+      const pngP = view.toImageURL('png', 2);
+      pngP.then((url) => {
+        downloadUrl(url, `${sensor.name} in ${'Unknown'}.png`);
```
sgratzl, line 89: since the detail view is showing one or more regions, can you add them to the file name instead of the 'Unknown'?

src/components/Vega.svelte
```diff
@@ -81,6 +81,8 @@
   $: updateSpec(patchedSpec);
   $: updateSignals(vegaPromise, signals);
 
+  export const vegaAccessor = () => vegaPromise.then((v) => v.view); // should be exported.
```
dlaliberte: I added the comment because I wasn't seeing "export" in the web page (due to compiler?), and not seeing changes at all, which turned out to be due to brokenness in the Emacs mode(s) so that changes were not being saved.   Deleted this comment in subsequent commit.

sgratzl (changes requested):

src/components/DetailView/DetailView.svelte
dlaliberte: Right... I forgot that I had simplified this because I wasn't able to get the names for some reason.  I'll figure that out.

dlaliberte: It's not clear how I can get the region names.  Your previous code generated the name with:
`${sensor.name} in ${$currentRegionInfo ? $currentRegionInfo.displayName : 'Unknown'}.png`
 but currentRegionInfo is not available in the version I started from.  Adding it to the imports and using the same string seems to work.

sgratzl: thx. while this will work for the single region case, it won't create a good title for a comparison case. Can you take a look how the h4 title tag a few lines below is generated. It should give you some hints to handle both cases in a generic way.

see also https://deploy-preview-556--cmu-delphi-covidcast.netlify.app/?sensor=doctor-visits-smoothed_adj_cli&level=county&date=20201007&signalType=value&encoding=color&mode=overview&region=42003&compare=37111

dlaliberte: OK, I should have noticed that.  I made the change, staged, committed, and pushed it, but I don't see it here yet.

dlaliberte: Hmm, I added a comment here that seems to have disappeared.  It said I made the change to the filename, staged, committed, and pushed it, but I don't see any evidence of it here.

dlaliberte: Ah, there is my earlier comment now.  And maybe the push didn't happen because ... the VSCode UI is very confusing, but I see it now.

sgratzl (approved):

src/components/DetailView/DetailView.svelte
sgratzl: Github notices you that it is now an outdated conversation on the code: 
image. 

The idea is that when you provide a review to a certain line of code but the line changes in a later commit it will remain there but warning you that the code you are viewing is outdated.

RoniRos:
I love the download button!  
Currently, it generates an image of just the net graph, for modular inclusion anywhere.
Do we want to consider also a more complete image-generation option? It could include the title, as well as a built-in acknowledgement of the source, e.g. a small font line at the bottom with something like:
"Data and image from Delphi Group's COVIDcast project, covidcast.cmu.edu"

dlaliberte:
I made a subsequent change that moves the title and subtitle into the chart, and moves the download and info buttons up to near the close button.  I can make another change to add an acknowledgement, though only displaying it in the generated image is probably more complex.

PR #558, opened 2020-10-09
feat: update population data based on indicator data

### LINKED ISSUE #553: Propagate updated population figures from indicators into www-covidcast
**Actual Behavior:**

Population in NYC boroughs is set to 1 in the map due to an old exception in JHU handling. We now use the combo source for cases & deaths, which uses USAFacts for NYC, and does not require this exception; JHU has additionally updated their geo policies and doesn't need the exception either.

1. go to https://covidcast.cmu.edu/index.html?sensor=indicator-combination-confirmed_7dav_incidence_num&level=county&date=20201006&signalType=value&encoding=color&mode=overview&region=36081
2. Mouse over Queens County

**Expected behavior**

Population: 2253858

**Screenshots**

image
image

**Rating Scale (1 is minor, 7 is severe):** 2? it's wrong, but the only place it shows up is the mouseover -- the prop figures come from the API, so they aren't affected.

**Additional context**

(used by combo cases/deaths for most locations) Population file used by USAFacts: https://github.com/cmu-delphi/covidcast-indicators/blob/main/usafacts/static/fips_population.csv

(used by combo cases/deaths for Puerto Rico) Population file used by JHU:https://github.com/cmu-delphi/covidcast-indicators/commit/6fda1b21a227dcda773f43b448da6060ba161bf7

Population file used by map (counties): https://github.com/cmu-delphi/www-covidcast/blob/main/src/maps/processed/county.csv

sgratzl:
closes #553

**Prerequisites**:

- [x] Build is successful

### Summary

updates the population with the latest file as mentioned in the issue when processing the raw data:

image


src/maps/process.js
```diff
@@ -110,7 +112,22 @@ function msa(level = 'msa') {
   return geo;
 }
 
-function counties(level = 'county') {
+async function generatePopulationLookup() {
+  const csv = csvParse(
+    await fetch(
+      'https://raw.githubusercontent.com/cmu-delphi/covidcast-indicators/main/_delphi_utils_python/delphi_utils/data/fips_pop.csv',
```
tildechris, line 118: Are there any cross-site considerations to be made when pulling this data from github directly?  Would it be better to pull the data down and package it as part of the site at build time?

tildechris (changes requested):
Approach LGTM, but should we incorporate fips_pop.csv as a site resource?

src/maps/process.js
sgratzl: the raw data is just processed on demand. It will generate among others the CSV file that is also checked in.

tildechris: Ah so, this is only at build time?  That makes sense.  It'd be nice if there was a npm package that we could pull down as a first-class dependency, but until then this make sense.

tildechris (approved):
LGTM.

PR #583, opened 2020-10-22
proof of concept: rt.live data as signal

sgratzl:
proof of concept using the rt.live data as an additional signal:

image


tildechris:
We've decided as a group that adding a R(t) signal, and/or integrating with rt.live isn't on our roadmap, so let's close this experiment.

PR #588, opened 2020-10-25
Dlaliberte/#586

### LINKED ISSUE #586: Tooltips flicker off and on when mouse moves across gaps between hovered values
**Actual Behavior:**

When I move the mouse horizontally over a chart and see the tooltip for some values, it flickers off and on again in many areas especially with bumpier data.  There appear to be gaps between hovered data values but these gaps are not due to missing data.

The charts for the small multiples and for Region Details demonstrate this problem.  This is mostly an annoyance, but it is also difficult (or impossible?) to see the tooltip for the data where these gaps are.

**Expected behavior**

I'd expect to the tooltip to either remain visible for the last hovered data point (if not too far away) or, better, select the data point that is vertically aligned with the mouse.

Annoying/distracting: 3

dlaliberte:
closes #586

**Prerequisites**:

- [x] Build is successful

### Summary

Just adds encodings: ['x'] to the selections, along with nearest: true, so that hovering finds the nearest x value, and this avoids the flickering tooltips as well.


dlaliberte:
I see now that the DetailView shows isolated points because it shows all points.  That's one solution.  Was this a workaround? Figuring out how to detect the isolated points was interesting; it seems to be fast as well.  Is it preferable to only show isolated points as I implemented in this change?  Maybe highlight each point as the user hovers?

sgratzl (commented):
the synchronized highlighting doesn't work for me anymore:

image

I would expect a vertical gray line in all charts for the current highlighted date as we had it before

sgratzl:
> I see now that the DetailView shows isolated points because it shows all points. That's one solution. Was this a workaround?

as far as I remember it was a design choice, since we had so much horizontal space and the user has the ability to zoom it, it was useful to highlight the data points in the line itself. In the small multiples it was too cluttered when I enabled the points

src/modes/overview/vegaSpec.js
```diff
@@ -127,6 +127,17 @@ export function createSpec(sensor, selections, dateRange, valuePatch) {
         calculate: 'datum.value == null ? null : datum.value / 100',
         as: 'pValue',
       },
+      {
```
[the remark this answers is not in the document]
dlaliberte: I tried the strokeCap trick, but the isolated points are no bigger than the line, which seemed too small.

src/modes/overview/vegaSpec.js
```diff
@@ -159,7 +170,7 @@ export function createSpec(sensor, selections, dateRange, valuePatch) {
             field: yField,
             type: 'quantitative',
             scale: {
-              domainMin: valuePatch && valuePatch.domain ? scalePercent(valuePatch.domain[0]) : 0,
+              domainMin: valuePatch && valuePatch.domain ? scalePercent(valuePatch.domain[0]) : undefined,
```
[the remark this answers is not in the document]
dlaliberte: That was probably fine.  I left this change in accidentally, since I was trying to get rid of showing the tooltip (showing 0 as the value) for undefined values.

dlaliberte:
> I would expect a vertical gray line in all charts for the current highlighted date as we had it before

You mean the red line?  I'm not sure what part of my change causes to be missing for the charts with no data, but it was unintended.  I'll see if I can get it back.

sgratzl:
> You mean the red line? I'm not sure what part of my change causes to be missing for the charts with no data, but it was unintended. I'll see if I can get it back.

no the synchronized grey line that is highlighted the hovered date in one chart in all the others:

e.g., in the current dev:

image

dlaliberte:
> > I see now that the DetailView shows isolated points because it shows all points. That's one solution. Was this a workaround?
> 
> as far as I remember it was a design choice, since we had so much horizontal space and the user has the ability to zoom it, it was useful to highlight the data points in the line itself. In the small multiples it was too cluttered when I enabled the points

Yes, it seems OK to show all the points in the DetailView, and yes, I can see that it would be too cluttered to show every point individually in the small multiples.  But showing only the isolated points that way seems advantageous, especially in cases where there are a lot of isolated points that would otherwise be invisible.  But at least the isolated points with the strokeCap trick makes them visible.

dlaliberte:
> > You mean the red line? I'm not sure what part of my change causes to be missing for the charts with no data, but it was unintended. I'll see if I can get it back.
> 
> no the synchronized grey line that is highlighted the hovered date in one chart in all the others:

Oh, I never noticed that before.  I bet it was because I took over what appeared to be an extra gray dot.  I can get that back, but maybe all my changes for showing isolated points should just be reverted (uncommitted).

The first commit, fixing the jittering toolbox, was the main thing I was trying to do.

src/modes/overview/vegaSpec.js
```diff
@@ -178,11 +178,12 @@ export function createSpec(sensor, selections, dateRange, valuePatch) {
             type: 'single',
             empty: 'none',
             on: 'mouseover',
-            nearest: false,
+            nearest: true,
+            encodings: ['x'],
```
dlaliberte, line 182: Sadly, it is exactly this addition of `encodings: ['x']` that interferes with showing the same gray line rule (and dot) in every chart.   This is the same change that makes the `nearest: true` setting be fast, since it avoids the voronoi computation, and it also stops the flickering of the tooltips, so it would be nice to be able to do this. 

I don't understand how all the other charts get this hover event, however.  Some magic is going on that I am not aware of.

dlaliberte (commented):
Do I need to "Submit review" to send other comments?  I don't think so, but who knows...

src/modes/overview/vegaSpec.js
sgratzl: if you change `resolveHighlightedTimeValue` to

```js
export function resolveHighlightedTimeValue(e) {
  const highlighted = e.detail.value;
  if (highlighted && Array.isArray(highlighted.date_value) && highlighted.date_value.length > 0) {
    return Number.parseInt(formatAPITime(highlighted.date_value[0]), 10);
  }
  return null;
}

then it should consider the selection x only properly

Comment (dlaliberte) [reply src/modes/overview/vegaSpec.js:182]: It does work with your resolveHighlightedTimeValue, thanks.  I was just starting to look at that function.  It does also work on the Region Details page, but not on the top-10 page.  I see there is a different onHighlight handler for the top-10 page so I'll see if I can make the appropriate changes there.

Comment (dlaliberte) [reply src/modes/overview/vegaSpec.js:179]: I was able to fix the top-10 page as well.  And also fixed the warning about finding 'nearest' 'line' in the range selector by moving the highlight2 selection to the 'rule' layer.

**Comment by dlaliberte** [comment]
I think everything is working pretty well now.  There is more work to do on the tooltip content, when multiple series are displayed in the same chart, but that is separate work.

File: src/components/tooltipUtils.js [hunk 4]
```diff
@@ -86,6 +86,10 @@ export function createVegaTooltipAdapter(svelteComponent, initialExtraProps = {}
       return;
     }
 
+    // If the datum value is null, no tooltip is needed.
```
sgratzl, line 89: should this condition be close to the hide condition on top such that the toolitp is hidden properly?

src/modes/overview/vegaSpec.js
```diff
@@ -178,11 +175,12 @@ export function createSpec(sensor, selections, dateRange, valuePatch) {
...
             clear: 'mouseout',
           },
         },
-        // use vertical rule for selection, since nearest is a real performance bummer
+        // use vertical rule for selection, since nearest point is a real performance bummer
         mark: {
```
sgratzl, line 184: do we still need this layer, since nearest is enabled or can we merge the selection back to the point layer below?

src/modes/top10/Top10.svelte
```diff
@@ -223,16 +223,19 @@
     highlightTimeValue = value;
   }, 10);
 
-  function onHighlight(e) {
+  function resolveHighlightedTimeValue(e) {
```
sgratzl, line 226: can you reuse the resolve function from the other file by importing it?

src/components/tooltipUtils.js
dlaliberte: I didn't need to hide it (by calling hide()?) in order to avoid showing it.  I do need to return before the following update() call, or else it would show an empty tooltip.

But I can move this up and combine it with the earlier condition.  Maybe the hide() call is needed if there is already an open tooltip?

src/modes/overview/vegaSpec.js
dlaliberte: We do still need the 'highlight' selection, and merging it with the following 'point' mark layer doesn't work unless you are positioned right where the point would be, since the selection is based on the mark itself, so the invisible vertical rule is what is being selected currently.

It also doesn't work to move the selection down to the following 'rule' mark, though I am not sure why.  Maybe order of the layers matters?

src/modes/top10/Top10.svelte
dlaliberte: Works.

sgratzl (approved):
👍

PR #600, opened 2020-10-29
survey dashboard

sgratzl:
applied most of the latest design changes: 

image

image

google doc for editing the questions: https://docs.google.com/document/d/148mmPJ6wFYdINA5EGW4Q4Tt7zZKR0qPQKMDdalJ7vlA


kateharwood:
On first glance this looks like a lot of text. 

I get that we want to give the salient info in an easy to digest way without just giving just that doesn't tell users much at a glance, but this looks kind of overwhelming as well. I wonder if adding some colors to the main numbers and "Steady"/"Increasing" words would help to focus people's attention quickly. Examples from NYTimes:
Screen Shot 2020-10-29 at 10 54 13 AM
Screen Shot 2020-10-29 at 10 53 53 AM

Even just putting some of the salient terms/numbers in the "CMU/Delphi Red" so we're not introducing new colors to the site would make this a bit more readable at a glance.


Another thought is to not include the descriptions of the signals at first, and have that be something users could expand beneath the titles, but I also understand why we might not want to do that.

philmcguinn:
I think the use of color helps here as well, here's an example from Google's Material guidelines to add to the pile
Screen Shot 2020-11-10 at 10 32 19 AM

I also noticed that the hover is a bit tricky, it seems like you need to get your mouse pointer on the exact line of the graph to see the hover tooltip. That's a diversion from our other charts that show data in a vertical line no matter where you hover on the chart.

sgratzl:
@tildechris @Letsgohyde what is the plan/design regarding this one?

tildechris:
This is looking really good.  Notes: 

* With your top-level metrics, you can delete: ", and about 70,000 people in the United States participate every day."
* Can you think of a way to more explicitly tie the lines in the graph to the metrics?  It isn't clear at first what the gray lines are
* What do you think about adding percentage change to:
  * The "high" category: What percentage is the current value from the high/low?  We'd have to play with the layout here to make room
  * The mouse-over box: This seems to be easier.  Can we add a percentage change to current date in the mouse-over box?  Then it's easy to see % change for any arbitrary date (if you have a mouse)

sgratzl:
> With your top-level metrics, you can delete: ", and about 70,000 people in the United States participate every day."

see google doc

> What do you think about adding percentage change to:
> 
>     * The "high" category: What percentage is the current value from the high/low?  We'd have to play with the layout here to make room
> 
>     * The mouse-over box: This seems to be easier.  Can we add a percentage change to current date in the mouse-over box?  Then it's easy to see % change for any arbitrary date (if you have a mouse)

I'm not sure how confusing it gets with different bases:
 * historical trend: 100% = reference value (a week ago) -> change = (current - ref) / ref
 * high category: 100% = high/low value -> change = (current - high) / high
 * tooltip: 100% = current value -> change = (hovered value - current) / current
atm we have: historical trend -> percentage current date compared to the value

tildechris (approved):
This is great work.  I'm excited to get this out there.  LGTM.

PR #604, opened 2020-10-29
add testing environment

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary

 * adds https://jestjs.io/ for unit-testing together with https://testing-library.com/docs/svelte-testing-library/intro -> see .spec files in src directory
 * adds https://www.cypress.io/ for e2e testing, see cypress/integrations spec files for examples
 * small refactoring for easier testing 

usage -> see README section


dlaliberte:
I reviewed the jest and cypress web sites and docs, and it all looks good to me.  The main question I would have is whether they work well with Svelte, given it's particular design characteristics.

sgratzl:
for cypress it doesn't make a difference since it is working with the website. for Jest, I use https://testing-library.com/docs/svelte-testing-library/intro which should deal with Svelte characteristics

tildechris (commented):
The testing framework here LGTM, but there's a lot to review.  

I think I understand the test hooks and writing tests, but I do not understand the changes around `msa.csv.js` and friends.  Can you summarize?

sgratzl:
> I think I understand the test hooks and writing tests, but I do not understand the changes around `msa.csv.js` and friends. Can you summarize?

importing non JavaScript files (like CSV) in JavaScript files usually requires a bundler like webpack or rollup with a custom loader. However, now we also have Jest in the loop which compiles the test code and thus also needs to handle to those kind of imports files. To avoid a more complicated test setup, I wrapped the CSV files to JavaScript modules, similar to what the current raw-loader of webpack was doing before.

tildechris (approved):
That sounds reasonable.  LGTM, thanks for the work here.

PR #608, opened 2020-10-30
Dlaliberte/#607

### LINKED ISSUE #607: Tables in "Region Details" tab flicker with each mouse move
**Actual Behavior:**

As you move your mouse across one of the charts horizontally, the date and other values flicker in the table above the chart.  Each move is redrawing the data for the 'current date' (where the red line is) and then for the date where the mouse is.  The flickering is particularly noticeable when the length of the current date and the mouse position date have a different length.

**Expected behavior**

The display should not flicker between two modes while the mouse is still within the chart.

**Additional context**

When the mouse leaves the chart, the data for the current date is displayed.  That is correct.  And the reason for the flicker is related to this final display for the current date. The `clear: 'mouseout'` in the vega spec means that each motion from one date to the next first sends an empty date for the clear, and then sends the date for the new position.  Maybe debouncing these signals will fix it.

dlaliberte:
closes #607

**Prerequisites**:

- [x] Build is successful

### Summary

Avoid the flickers between mouseout and mouseover events by debouncing.  This usually ignores the mouseout, unless there is no mouseover that follows.

### DIFF
--- a/src/modes/single/SingleLocation.svelte
+++ b/src/modes/single/SingleLocation.svelte
@@ -23,7 +23,6 @@
   // achieves that goal, at least most of the time.
   const debouncedHighlightTime = debounce(
     (value) => {
-      console.info('highlighted valuu', value);
       highlightTimeValue = value;
     },
     1,
--- a/src/modes/single/SingleLocation.svelte
+++ b/src/modes/single/SingleLocation.svelte
@@ -11,7 +11,7 @@
   } from '../../stores';
   import { sensorList } from '../../stores/constants';
   import SensorCard from './SensorCard.svelte';
-  import debounce from 'lodash-es/throttle';
+  import debounce from 'lodash-es/debounce';
   import { selectionColors } from '../../theme';
   import { resolveHighlightedTimeValue } from '../overview/vegaSpec';
 
--- a/src/modes/single/SingleLocation.svelte
+++ b/src/modes/single/SingleLocation.svelte
@@ -11,20 +11,29 @@
   } from '../../stores';
   import { sensorList } from '../../stores/constants';
   import SensorCard from './SensorCard.svelte';
-  import throttle from 'lodash-es/throttle';
+  import debounce from 'lodash-es/throttle';
   import { selectionColors } from '../../theme';
   import { resolveHighlightedTimeValue } from '../overview/vegaSpec';
 
   let highlightTimeValue = null;
 
-  const throttled = throttle((value) => {
-    highlightTimeValue = value;
-  }, 25);
+  // Each mouse move across the chart results in a mouseout for previous highlight
+  // date immediately followed by mouseover for the next date, if any.  We need to
+  // ignore the first event unless there is no second.  A very short debounce time
+  // achieves that goal, at least most of the time.
+  const debouncedHighlightTime = debounce(
+    (value) => {
+      console.info('highlighted valuu', value);
+      highlightTimeValue = value;
+    },
+    1,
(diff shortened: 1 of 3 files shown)


src/modes/single/SingleLocation.svelte
```diff
@@ -11,20 +11,29 @@
...
-  const throttled = throttle((value) => {
-    highlightTimeValue = value;
-  }, 25);
+  // Each mouse move across the chart results in a mouseout for previous highlight
+  // date immediately followed by mouseover for the next date, if any.  We need to
+  // ignore the first event unless there is no second.  A very short debounce time
+  // achieves that goal, at least most of the time.
+  const debouncedHighlightTime = debounce(
+    (value) => {
+      console.info('highlighted valuu', value);
```
sgratzl, line 26: can you remove the logging statement, please

src/modes/single/SingleLocation.svelte
```diff
@@ -11,20 +11,29 @@
...
   let highlightTimeValue = null;
 
-  const throttled = throttle((value) => {
-    highlightTimeValue = value;
-  }, 25);
+  // Each mouse move across the chart results in a mouseout for previous highlight
+  // date immediately followed by mouseover for the next date, if any.  We need to
+  // ignore the first event unless there is no second.  A very short debounce time
+  // achieves that goal, at least most of the time.
+  const debouncedHighlightTime = debounce(
```
sgratzl, line 24: why is a debouncing better than a throttling?

src/modes/single/SingleLocation.svelte
dlaliberte: Done, but forgot to push it.

src/modes/single/SingleLocation.svelte
dlaliberte: Throttling will reduce the rate that it lets events through, so it might pick the first (clear) event or the second event (for the mouse over a new date).  We only want the second one of each pair, if there is a second one, otherwise we want the first one.  Debouncing throws away extra events that arrive too quickly, which are the 'bounces' when you really do have bouncing events.  In this situation, the two events are very close, so we can have the debouncing always throw away the first event.

src/modes/single/SingleLocation.svelte
sgratzl: please push it

dlaliberte: OK, I thought I did it, but must have been in another branch.  Did it again.

src/modes/single/SingleLocation.svelte
```diff
@@ -11,20 +11,28 @@
   } from '../../stores';
   import { sensorList } from '../../stores/constants';
   import SensorCard from './SensorCard.svelte';
-  import throttle from 'lodash-es/throttle';
+  import debounce from 'lodash-es/throttle';
```
sgratzl, line 14: I just notice that: you call it debounce but import throttle? was this intentional?

dlaliberte: It was unintentional in the sense that I assumed debounce was part of throttle, and mysteriously, it worked.  However, I see there is a separate debounce.js, so changing the dependency to that.  This resulted in vscode complaining "Could not find declaration file for module 'lodash-es/debounce'" and the suggestion to do npm install @types/lodash-es, which I did, and it changed a couple of json config files.  Those changes are now pushed.  Let me know if this is not right.

sgratzl (approved):
👍

PR #614, opened 2020-11-02
Dlaliberte/#589

### LINKED ISSUE #589: Top 10 time series plots get truncated
**Actual behavior:**

Time series plots for "Top 10" view get (erroneously) truncated on the y-axis.

**Steps to reproduce the bug:**

https://covidcast.cmu.edu/?sensor=doctor-visits-smoothed_adj_cli&level=county&date=20201019&signalType=value&encoding=color&mode=top10&region=42003

**Expected behavior:**

The y-axis should scale so that the dynamic range of the time series is contained within the plot (as it does in the small multiple time series panel, or the big time series panel, in the "Map Overview" view).

**Screenshots:**

Truncated time series plots in "Top 10" view:

Screen Shot 2020-10-26 at 12 53 55 PM

Here's what the first one looks like in the small multiples time sets panel of the "Map Overview" view:

Screen Shot 2020-10-26 at 12 59 19 PM

**Rating scale (1 is minor, 7 is severe):**

2 or 3

dlaliberte:
closes #589 

**Prerequisites**:

- [x] Build is successful

### Summary

Render line chart with clipped regions.


src/modes/overview/vegaSpec.js
```diff
@@ -121,10 +126,49 @@ export function createSpec(sensor, selections, dateRange, valuePatch) {
     transform: [
       ...(sensor.hasStdErr ? (isPercentage ? stdErrTransformPercent : stdErrTransform) : []),
       {
-        calculate: 'datum.value == null ? null : datum.value / 100',
         as: 'pValue',
+        calculate: 'datum.value == null ? null : datum.value / 100',
+      },
+      {
+        as: 'clipped',
+        calculate: !clipping + '|| datum.' + yField + ' == null ? false : datum.' + yField + ' > ' + yMaxScaled,
```
sgratzl, line 134: consider using backticks (``) for better readability of string interpolation

src/modes/overview/vegaSpec.js
```diff
@@ -121,10 +126,49 @@ export function createSpec(sensor, selections, dateRange, valuePatch) {
...
+        calculate: 'datum.clipped == null ? false : !datum.clipped',
+      },
+      {
+        as: 'clippedData',
+        calculate: 'datum.clipped ? datum.' + yField + ' : null',
+      },
+      {
+        window: [
+          {
+            as: 'previousNotClipped',
```
sgratzl, line 147: I'm wondering whether it should be the first value that is clipped (first_value) or the first before it is clipped (lag).

src/modes/top10/Top10.svelte
```diff
@@ -174,15 +175,44 @@
     return data;
   });
 
+  // Compute the neighborhood around the current date.
```
sgratzl, line 178: I'm not sure whether is as reactive as it should be. 

to my knowledge the `const date` won't be reactive, the `$: neighboorhood` would be but just uses non-reactive variables. 

Have you considered extracting this whole logic to a single function as in:

```js
$: neighborhood = computeNeighborhood($currentDateObject, $smallMultipleTimeSpan);
```

sgratzl: regarding the neighboorhood computation:

https://github.com/cmu-delphi/www-covidcast/blob/c150bb3ca729ae618b927363d7187019499cde6f/src/stores/index.js#L124-L134

computes the small multiple time span which limits the view to four months around the current date. As far as I understand the neighborhood does a similar thing but for around 1 month. Can you extract the logic of both function into a utility function having a date and the offset as parameter and it will return the date range.

src/modes/overview/vegaSpec.js
```diff
@@ -103,16 +103,21 @@ const stdErrTransformPercent = [
  */
 export function createSpec(sensor, selections, dateRange, valuePatch) {
   const isPercentage = sensor.format === 'percent';
+  const scalePercent = isPercentage ? (v) => v / 100 : (v) => v;
+
+  const clipping = valuePatch && valuePatch.domain ? true : false;
```
sgratzl, line 108: as far as I understand if `clipping` is false a bunch of layers and calculations are not used. For performance reasons (e.g., in the small multiple case), one could consider removing (or better not adding) them.

src/modes/overview/vegaSpec.js
dlaliberte: Yes, it seemed easier while I was experimenting with many variations to just use concatenation, but actually, backticks everywhere would have been easier.

src/modes/overview/vegaSpec.js
dlaliberte: The clipped region should start and end with clipped values so that the region is completely flat.  That's why we want to determine whether the current value is clipped and the previous value is not clipped.

To compute previousNotClipped we look at the previous value (lag of 1) to check if it is not clipped, so that this will be the start of the clipped region, but only if the current value _is_ clipped.  I would collapse logic into fewer steps if I could figure out how.

For the end of the clipped region, the logic is similar, but the default lead (1) seemed to be 1 too far, hence the param: 0.

This is all fairly confusing.  I'll add a comment about it.

src/modes/top10/Top10.svelte
dlaliberte: > I'm not sure whether is as reactive as it should be.
> 
> to my knowledge the `const date` won't be reactive, the `$: neighboorhood` would be but just uses non-reactive variables.

That sounds right.  I'm not clear yet on what needs to be reactive.   But since the $currentDateObject is never changed in the top-10 page, the neighborhood would also not need to be reactive.  However, it seems like we _should_ allow the $currentDateObject to be changed.   

> 
> Have you considered extracting this whole logic to a single function as in:
> 
> ```js
> $: neighborhood = computeNeighborhood($currentDateObject, $smallMultipleTimeSpan);
> ```

That sounds like a good idea.  Although $smallMultipleTimeSpan turns out to not be precisely relevant on the top-10 page since it is a subset of the top-10 timespan.

dlaliberte: > regarding the neighboorhood computation:  [...]
> 
> computes the small multiple time span which limits the view to four months around the current date. As far as I understand the neighborhood does a similar thing but for around 1 month. 

Well, in prior discussions, we talked about a neighborhood of a couple weeks or months, but it seems more general and useful to use a neighborhood that is a fraction of the top-10 timespan, however long that is.  That way, if we allow zooming or panning, or other ways of selecting the total timespan, the neighborhood should always scale to the current timespan.

But I couldn't find a way to determine the top-10 timespan, other than by iterating through potentially all the data, given that some data could be missing.   Is there a way?  Alternatively, we could _always_ use the $smallMultipleTimeSpan, which has the advantage of always being consistent with what we show in small multiples in other places.

> Can you extract the logic of both function into a utility function having a date and the offset as parameter and it will return the date range.

This sounds like it would be a useful function in any event, though there is a slightly different purpose here.  The small multiple time span is always equally 1 month around the current date, or shifted back if needed.  But this neighborhood calculation was intended to be a subset of whatever the current data timespan is, so we may have to shift back (if overlapping the end of the timespan) or forward (if overlapping the start of the timespan).  I'll see if I can generalize this, however.

src/modes/overview/vegaSpec.js
dlaliberte: Yes, it would be good to avoid computing all the extra data and layers if not needed.  That should be doable.

dlaliberte:
Here's what my current (still local) version looks like:

image

src/util.js
```diff
@@ -23,6 +23,38 @@ export function transparent(colors, opacity) {
...
+ * @param {Date} date
+ * @param {Date} startDate
+ * @param {Date} endDate
+ * @param {number|null} duration of neighborhood, in milliseconds
+ */
+export function computeNeighborhood(date, startDate, endDate, duration) {
+  if (duration == null) {
+    duration = endDate - startDate;
+  }
+  const neighborhood = { start: date - duration / 2, end: date.getTime() + duration / 2 };
```
sgratzl, line 39: why do you call `getTime` in one but not in the other case? I think JavaScript will coalesce also the first case but the second case is more clear

sgratzl (approved):
code looks good to me, just a small comment. Cannot comment much on the overall design

src/util.js
dlaliberte: The getTime (or some cast to get milliseconds) is necessary because of the following '+' operator, which gets interpreted as string concatenation.   I tried duration / 2 + date, but that doesn't work either.  However, Number(date) + duration/2 does work.    It would be more consistent to use getTime() or Number() always, but that's annoying too.

Also, should this function return a Date instead of milliseconds?

sgratzl: I get the reasoning behind `getTime`, what is confusing me is the inconsistency that when computing `start` you don't call `.getTime()` but when computing `end` you do.

dlaliberte: Ah, because the '-' operator is only numeric, so the date is auto converted to the number of milliseconds.

tildechris:
@dlaliberte can you note good testing parameters for geo area and date that demonstrates this behavior and solicit feedback on slack.?

dlaliberte:
Here is a simple case, at least for today.  Visit https://deploy-preview-614--cmu-delphi-covidcast.netlify.app/?sensor=hospital-admissions-smoothed_adj_covid19_from_claims&level=county&date=20201107&signalType=value&encoding=color&mode=top10&region=42003
Then click on the "Name" header twice to get reverse order with York on top.  You'll see a few instances of clipping.

philmcguinn:
I like the new visual style, looks clear in the example you posted above.

I do wonder if our parameters for sizing are too strict though. What I mean is that we seem to be pretty aggressive in re-scaling, where things are being chopped that are only 1-2% out of the visual range. In the following example it looks to be about 3% above. I would expect clipping to be used in more extreme circumstances than something like this.

Screen Shot 2020-11-12 at 3 24 09 PM

If we decide to loosen the clipping requirements a bit, what levers do we have?

dlaliberte:
The clipping threshold is nothing more than the maximum value across about 1/2 of the visible time range, split evenly if possible around the current date (the red line).  So anything above that threshold in the rest of the charts is clipped, regardless whether it is a lot or a little.

I was also thinking it might be nicer to look at how much the threshold would have to increase to avoid some clipping.  But this is actually a more complex and interesting problem than it sounds because of the bumpiness of the data.  We would have to decide on some arbitrary cut-off, say increase the threshold no more than 10%, at which point there might be some other peak that is just beyond that threshold.  It would be possible to sort all the data values and look at the vertical distribution, but the horizontal distribution also matters.    I don't have any good ideas at the moment, but I will be thinking about it.

PR #617, opened 2020-11-05
Use correct field when computing max of values.

### LINKED ISSUE #615: Top 10 list refreshes multiple times, occasionally resulting in incorrect clipping
**Actual Behavior:**

The top 10 list refreshes the display of the charts multiple times, partly due to dynamic loading of data but also resize events and clipping recalculations.  This is visible in origin/dev versions, especially after #575.  See https://deploy-preview-575--cmu-delphi-covidcast.netlify.app

The problem is that, occasionally and inconsistently, the resulting display clips the lines incorrectly.  

Go to the Top 10 list, and keep switching the indicator that is 'Displaying' by using the drop-down selector until you notice that the clipping is incorrect.

The 7-day average charts for 'Cases' and 'Deaths' are always clipping incorrectly, but for a different reason.

**Expected behavior**

Refreshes should be minimized.
Clipping should be consistently correct.

5

dlaliberte:
partly fixes #615

**Prerequisites**:

- [x] Build is successful

### Summary

Instead of always using 'value', use the primaryField (with 'pValue' replaced with 'value').
Do this for the main sensor and 'others'.
Also globalizes debouncedResize, to reduce refreshes a bit.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -4,7 +4,7 @@ module.exports = {
     '^.+\\.js$': 'babel-jest',
   },
   // // transform all files including node modules
-  transformIgnorePatterns: ['node_modules/(?!(svelte|svelte-icons)/)'],
+  transformIgnorePatterns: ['node_modules/(?!(svelte|svelte-icons|lodash-es)/)'],
   moduleFileExtensions: ['js', 'svelte'],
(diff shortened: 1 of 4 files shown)


sgratzl (approved):
👍

PR #627, opened 2020-11-11
proof of concept: wrapping www-covidcast with hugo

sgratzl:
image


themes/delphi/assets/uikit/package.json
```diff
@@ -0,0 +1,90 @@
+{
```
tildechris, line 1: Is there a reason all of these uikit files are included in this draft?  If they need to live outside of node_modules, there should be a script to move them into the right place and then .gitignore them rigth?

sgratzl: that is the hugo theme 1:1 from the www-main website. In this example it is copied on purpose such that in a different version we can share the theme more easily

sgratzl:
closing since covidcast will be embedded using its assets and not an independent hugo page

PR #628, opened 2020-11-12
feat: limit add another to same geo level

### LINKED ISSUE #624: Add another geo area should be constrained to the current geo-type
Re: #619 

"While in map overview mode, I was on "state" mode for the geo type pull down menu, and then in compare mode, when I search for things to add, under "ADD ANOTHER", I get counties popping up. This is confusing. Actually, even when I entered compare mode, I already got "Allegheny County, PA" as the location I'm beginning with, for the comparisons. (This must be because it detected this as my home location, based on IP address?). Which is also confusing."

We should constrain the comparisons to only geo-types for the currently selected area (only show states for states etc.)

I'm not sure if I can think of a good use-case to heterogeneously compare across geo types.

sgratzl:
closes #624

**Prerequisites**:

- [x] Build is successful

### Summary

limits the possible items in the add another search box to more of the same level, e.g. here states:

image

### DIFF
--- a/src/components/Search.svelte
+++ b/src/components/Search.svelte
@@ -47,8 +47,8 @@
   /**
    * @type {any[]}
    */
-  export let selectedItems = [];
-  $: multiple = selectedItems.length > 0;
+  export let selectedItems = null;
+  $: multiple = selectedItems != null;
   export let maxSelections = Number.POSITIVE_INFINITY;
 
   let text;
@@ -98,7 +98,7 @@
     // store reference to the origial item
     item,
   }));
-  $: selectedLabelLookup = new Set(selectedItems.map((s) => labelFunction(s)));
+  $: selectedLabelLookup = new Set((selectedItems || []).map((s) => labelFunction(s)));
 
   function limitListItems(items) {
     if (maxItemsToShowInList <= 0 || items.length < maxItemsToShowInList) {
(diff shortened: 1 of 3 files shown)


dlaliberte:
I'm not seeing this work entirely.  If I first remove any counties, then when I switch to States, the completion list shows states, but otherwise, it continues to show counties.  Same thing in reverse: if any states have been selected, after switching to Counties, it still shows states.    If I have both counties and states selected, then counties override states in the decision of what to show.    Seems like it should always show whatever is chosen in the "for ..." selector regardless what is currently selected.

sgratzl:
atm the logic is that is shows more of the same. So, if you have selected a state (like Minnesota) it will list more states. If you have selected a county (like Allegheny County) it will list counties. There is no connection to what geo type one has selected in the map.

tildechris:
I like how this works.  I think I found a bug, however:

* Clear any selected region
* Go to "Region Details"
* Search and select "Allegheny County, PA"

The text Allegheny County is text repeated in the text field making it awkward to add the next region.

sgratzl:
> I like how this works. I think I found a bug, however:
> 
>     * Clear any selected region
> 
>     * Go to "Region Details"
> 
>     * Search and select "Allegheny County, PA"
> 
> 
> The text Allegheny County is text repeated in the text field making it awkward to add the next region.

should be fixed

tildechris (approved):
LGTM, thanks Sam.

PR #630, opened 2020-11-12
feat: restore focus upon dialog close

### LINKED ISSUE #598: Closing Modal does not return user focus to previous point in page
**Actual Behavior:**

When closing a modal pop-up via keyboard the user is not returned to their previous point in the page. Instead, they are moved elsewhere. For detail views it's returned to the top of the sensor list. For information, it's sent to just previous to the navigation. 

When I...

1. Go to https://covidcast.cmu.edu/index.html
2. Focus to the sensor list, navigate to the either to ? button or the Detail button
3. Hit enter, observe you're correctly moved to the new window
4. Hit enter on close button
5. Observee you're not back in the div order from where you came 

**Expected behavior**

When closing a modal that had grabbed the focus, it should return to where the user had previously been.

**Screenshots**

In the gif, first the detail view is opened and closed and the focus goes from the detail view back to the top of the sensors.  Then the info window is opened and closed and then the focus is tab advanced a few times, opening the navigation

covidcast-modal

3

**Context**

- Browser: Mac OS, Chrome

**Additional context**

Returning to the sensors isn't that bad, if a fix is to be prioritized it would be making the closing of the info window not jump to just before the nav

sgratzl:
closes #598

**Prerequisites**:

- [x] Build is successful

### Summary

restores the old focus upon closing a dialog which forces the focus on the close button.

focus

### DIFF
--- a/src/components/DetailView/DetailView.svelte
+++ b/src/components/DetailView/DetailView.svelte
@@ -74,17 +74,27 @@
   function onEscCheck(e) {
     if (e.key === 'Escape' || e.key === 'Esc') {
       trackEvent('detail-view', 'close', 'keyboard');
+      restoreFocus();
       dispatch('close');
     }
   }
 
   let close = null;
+  let oldFocus = null;
 
   onMount(() => {
     if (close) {
+      oldFocus = document.activeElement;
       close.focus();
     }
   });
+
+  function restoreFocus() {
+    if (oldFocus) {
+      oldFocus.focus();
+      oldFocus = null;
+    }
+  }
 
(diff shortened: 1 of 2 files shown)


dlaliberte (approved):
Looks good.  Works fine.   

Though it seems like this should be a feature built-in to a dialog widget.   Also, a weird edge case: it can only remember one previous oldFocus, so if you switch from the previously focused thing to yet another thing, it will still restore focus to the previously focused thing.

tildechris (approved):
LGTM.  Thanks Sam.

PR #633, opened 2020-11-13
improve detail view (axis) labels

### LINKED ISSUE #625: Y-axis label doesn't differentiate between counts and ratios
Re: #619 

> When I toggle between incidence/cumulative and counts/ratios (per 100,000) the y-label on the big time series plot remains the same. It should change. Certainly when you toggle between counts/ratios, it's no longer correct to label the y-axis "Cases" or "Deaths" for example, it should say "Cases per 100,000" or "Deaths per 100,000".

Fix the y-axis label so that we are being precise with the units that we are showing.

dlaliberte:
closes #625

**Prerequisites**:

- [x] Build is successful

### Summary

Compute the y-axis title based on the primaryValue.
Also adjust the left padding to add more space for cumulative counts and ratios, since the ticks are larger.
Also fix the chart title (since the first part of the title was not about the indicator in the chart)


dlaliberte:
For Incidence Ratios:

image

Similar for Cumulative Cases and Cumulative Cases per 100,000.

ryantibs:
Hi @dlaliberte, oops I didn't see this in time and left comments on issue #625.  See https://github.com/cmu-delphi/www-covidcast/issues/625#issuecomment-727142387

dlaliberte:
> Hi @dlaliberte, oops I didn't see this in time and left comments on the other issue, rather than doing a review. See [#625 (comment)](https://github.com/cmu-delphi/www-covidcast/issues/625#issuecomment-727142387)

No problem.  I'm seeing all comments everywhere (I believe :).

src/components/DetailView/vegaSpec.js
```diff
@@ -137,8 +144,8 @@ export function createSpec(sensor, primaryValue, selections, initialSelection, t
       contains: 'padding',
       resize: true,
     },
-    padding: { left: 50, right: 2, top: 45, bottom: 5 },
-    transform: sensor.hasStdErr ? stdErrTransform : [],
+    padding: { left: leftPadding, right: 2, top: 50, bottom: 5 },
+    transform: [...(sensor.hasStdErr ? stdErrTransform : [])],
```
dlaliberte, line 148: This used to have the array wrapper around spreading of the elements, but it got dropped at some point.  However, now that I think about it, they are equivalent, so I'll put it back to the way it was.

src/components/DetailView/vegaSpec.js
```diff
@@ -118,6 +118,13 @@ export function colorEncoding(selections) {
  * @param {Array<string>} title
  */
 export function createSpec(sensor, primaryValue, selections, initialSelection, title) {
+  const ratioSuffix =
```
sgratzl, line 121: an alternative strategy is using the `$signalCasesOrDeathOptions` (with `.ratio` and `.cumulative`) as way to differentiate variants

dlaliberte: That would be cleaner, but it's not working for me.

```
  const ratioSuffix = $signalCasesOrDeathOptions && $signalCasesOrDeathOptions.ratio ? ' per 100,000 people' : '';
  const yAxisTitle = sensor.yAxis + ratioSuffix;

  const leftPadding = $signalCasesOrDeathOptions && $signalCasesOrDeathOptions.cumulative ? 90 : 50;
```

sgratzl:
regarding the yAxis text. I wonder whether we should apply a similar strategy as it is being used for `mapTitleText` in the description. So, that in the google doc it is actually an object with the four combinations which will be converted in code into a function based on the signal options object. This would extract the text constants similar to the other parts and also avoid that have a specific logic in the detailview

sgratzl:
can you merge the latest dev into this PR, please

dlaliberte:
> can you merge the latest dev into this PR, please

Done, but I am seeing errors about sass-loader and uikit/dist/js/uikit-icons.js.  I'm guessing I have to install something, but I don't see any hint about what I need to do, as I have on other occasions.

sgratzl:
> Done, but I am seeing errors about sass-loader and uikit/dist/js/uikit-icons.js. I'm guessing I have to install something, but I don't see any hint about what I need to do, as I have on other occasions.

just make sure to run `npm install`

dlaliberte:
> just make sure to run `npm install`

That works.  Thanks.  Normally I only need to do npm start, so I forgot about npm install.

However, now it fails to save the date range all the time.  I'll investigate.

dlaliberte:
> However, now it fails to save the date range all the time. I'll investigate.

Oh, lol. I got my changes mixed up.  This one is not about the date ranges, just the axis labels and related text and alignment.

dlaliberte:
> regarding the yAxis text. I wonder whether we should apply a similar strategy as it is being used for `mapTitleText` in the description. So, that in the google doc it is actually an object with the four combinations which will be converted in code into a function based on the signal options object. This would extract the text constants similar to the other parts and also avoid that have a specific logic in the detailview

We should probably make the more comprehensive change you suggest, in which case, this PR could end up with rundant "per 100,000 people" added to the label.  We can undo that part of this PR later, or do the more comprehensive change before merging this PR.    What do you suggest?

sgratzl:
> 100,000 people" added to the label. We can undo that part of this PR later, or do the more comprehensive change before merging this PR. What do you suggest?

can you create a follow up issue for it, please and then we merge that and change it later

dlaliberte:
> can you create a follow up issue for it, please and then we merge that and change it later

Done: Comprehensive, consistent labeling of Cases and Deaths re incidence/ratios/cumulative #642

sgratzl (approved):
👍

PR #635, opened 2020-11-16
fix: date range initial selection was lost

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary

fix that the initial date range is selected again at the beginning.

image

the solution was somehow to define a "dummy" data such that the scale has always some date range to convert dates to pixels. As an alternative if you change `autoSize.type` back to `pad` you also don't have the problem (don't know why tho)


dlaliberte (approved):
Looks good.  Glad you found a fix.

PR #641, opened 2020-11-17
Preserve modified date range while DetailView is displayed.

### LINKED ISSUE #623: X-Range should persist across selection
Re: #619

When using the large time series chart, one can use the date range widget to change the range of the plot:

image

Currently if you add another area to the selection, this date range gets reset, causing the user to have to navigate back to the area they found interesting.

We should persist the current date range across geo selections to improve this behavior.

dlaliberte:
closes #623

**Prerequisites**:

- [x] Build is successful

### Summary

Preserve the modified date range in the DetailView between changes of sensor or regions. Also entering and exiting compare mode leave the date range as it is. Only closing the DetailView restores the default, next time it is opened.


dlaliberte:
It actually fails to preserve the date range the first time after opening the DetailView.  And if you select a sensor with no data, it loses the date range entirely.

dlaliberte:
Fixed the "no data" situation.  The failure to preserve the date on the first time doesn't happen if I am watching with the debugger, so there is a timing issue or something that I am not understanding.

dlaliberte:
Redefined the dateRange variable to use writable, but it still fails (most of the time) the on the first change, but works thereafter.  I'm stumped.

src/components/DetailView/DetailView.svelte
```diff
@@ -59,11 +65,42 @@
...
+  }
+
+  function onDateRangeChange(event) {
+    // All signal events are sent to all on:signal handlers (??), so check if this is the right handler.
+    if (event.detail.name !== 'dateRange') {
+      return;
+    }
+    const dr = event.detail.value && event.detail.value.date_value;
+    if (dr && Array.isArray(dr) && dr.length > 0) {
+      console.info('onDateRangeChange', dr);
```
sgratzl, line 88: can you remove the log statement, please

src/components/DetailView/DetailView.svelte
```diff
@@ -59,11 +65,42 @@
...
+      if (
+        (!Number.isNaN(dr[0]) && dr[0] !== $detailViewTimeSpan[0].getTime()) ||
+        (!Number.isNaN(dr[1]) && dr[1] !== $detailViewTimeSpan[1].getTime())
+      ) {
+        detailViewTimeSpan.set([new Date(dr[0]), new Date(dr[1])]);
+        console.info('onDateRangeChange new dates', $detailViewTimeSpan);
```
sgratzl, line 94: same here

sgratzl (commented):
> Redefined the dateRange variable to use writable, but it still fails (most of the time) the on the first change, but works thereafter. I'm stumped.

how can I reproduce that? can you make a screenshot showing the behavior?

src/components/DetailView/DetailView.svelte
dlaliberte: Yes, well, my thought was that it helps to have the log messages while trying to debug why it fails the first time.  I'll remove them before this is ready to merge.

src/components/DetailView/DetailView.svelte
dlaliberte: ack.  will do.  

BTW I don't think the console messages make the problem worse, though that might be possible.

src/components/DetailView/DetailView.svelte
sgratzl: then please convert this PR to a draft such that I know it is not ready to be reviewed/merged. To my understanding if I'm assigned as a reviewer and the PR is not a draft, I assume I should review it because it is in its final state

dlaliberte:
> how can I reproduce that? can you make a screenshot showing the behavior?

I'm not sure if I can do a screen recording.  Need a tool allowed by Google that works with macbook and/or chrome.

But reproducing this is straightforward.  On a fresh page (I click on "COVIDcast" link in header), 
- starting with Map Overview, 
- click on magnifier icon for COVID-Related Doctor Visits, 
- drag the date range slider back a couple months, 
- click the magnifier for COVID-like Symptoms
- Notice that the date range slider is reset to the small multiple default.
- Try it again (without restarting), and notice it works correctly thereafter.

Actually, the latest version preserves the change after closing the detail view as well, or going to any other view.  That's because it is storing it in detailViewTimeSpan, so it never gets reset to the default after that.  It only fails the very first time.

dlaliberte:
> Actually, the latest version preserves the change after closing the detail view as well, or going to any other view. That's because it is storing it in detailViewTimeSpan, so it never gets reset to the default after that. It only fails the very first time.

After merging in the latest from origin/dev, now it does work the first time as well.  So something has changed maybe about the timing of page loading or whatever, so I wonder if this will be reliable.

Also, the persistence of the date range change could be a useful feature, but now we have the opposite problem, that there is no way to reset to the default, other than refreshing the page.   I'll remove it from the store, and see if that works correctly now.

src/components/DetailView/DetailView.svelte
dlaliberte: All works correctly now.

src/components/DetailView/DetailView.svelte
dlaliberte: I needed your feedback (which was useful, by the way), and assumed you would want to see the console messages as well.  

But I think the 'draft' status is not quite right, unless I mean to say "you can ignore this PR until further notice", which is not what I wanted to convey.   We would need another status that indicates "Not ready to submit, but your feedback is needed".   And regardless, even if it is not in 'draft' mode, the PR would not be ready to submit until approved, so it means the same thing.

src/stores/index.js
```diff
@@ -133,6 +133,8 @@ export const smallMultipleTimeSpan = derived([currentDateObject], ([date]) => {
   return [min, max];
 });
 
+export const detailViewTimeSpan = writable([]);
```
sgratzl, line 136: do we still need that?

sgratzl (approved):
looks good beside the one comment regarding the store

src/stores/index.js
dlaliberte: Nope.  I forget to remove that.

src/components/DetailView/DetailView.svelte
```diff
@@ -59,11 +59,40 @@
...
+    // All signal events are sent to all on:signal handlers (??), so check if this is the right handler.
+    if (event.detail.name !== 'dateRange') {
+      return;
+    }
+    const dr = event.detail.value && event.detail.value.date_value;
+    if (dr && Array.isArray(dr) && dr.length > 0) {
+      if (
+        (!Number.isNaN(dr[0]) && dr[0] !== detailViewTimeSpan[0].getTime()) ||
+        (!Number.isNaN(dr[1]) && dr[1] !== detailViewTimeSpan[1].getTime())
+      ) {
```
dlaliberte, line 85: Maybe we shouldn't bother checking whether the value has changed, since it won't be used immediately anyway.

But the check that both values are not NaN is important anyway, since NaNs will be sent when there is no data for the sensor.

PR #643, opened 2020-11-18
Avoid flickering tooltips on small multiples 

### LINKED ISSUE #586: Tooltips flicker off and on when mouse moves across gaps between hovered values

dlaliberte:
closes #586

**Prerequisites**:

- [x] Build is successful

### Summary

This change uses the same debouncedHighlight that works on the Region Details page because it is also needed for the small multiples charts.  The code is moved to small multiples vegaSpec.js where it is reused for both.


dlaliberte:
I see I broke the cross-chart highlighting.... Working on it...

Also, there is a handy placement option for popper that could avoid most of the problem with the overlap.

sgratzl (approved):
👍

PR #616, opened 2020-11-05
UIKit Adaptations

sgratzl:
**Prerequisites**:

- [x] Unless it is a hotfix it should be merged against the `dev` branch
- [x] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [x] Build is successful
- [x] Code is cleaned up and formatted

### Summary

imports and applies the UIkit into covidcast: 

image
image
image

atm. the new version doesn't have a header since it is designed to be embedded into the main website, either as an iframe or directly via the scripts

PR #661, opened 2020-11-30
Improve layout and alignment of top-10 tab

### LINKED ISSUE #657: Wide windows create excessive whitespace between columns in top10
**Actual Behavior:**

In a wide window, the space between the Name and Population text in the Top-10 view is sufficiently greater than the space between two entries that it becomes difficult to identify rows.

1. go to https://dev--cmu-delphi-main.netlify.app/covidcast/?sensor=doctor-visits-smoothed_adj_cli&level=county&date=20201126&signalType=value&encoding=color&mode=top10&region=42003
2. make window wider

**Expected behavior**

Columns should have a maximum width to preserve legibility of rows.

**Screenshots**

image

**Rating Scale (1 is minor, 7 is severe):** 1

**Context**

- Browser: Chrome 87.0.4280.67 on a Mac

dlaliberte:
closes #657 

**Prerequisites**:

- [x] Build is successful

### Summary

Change the sizing and alignment of the top-10 table columns.  A max-width doesn't exactly work, but distributing the width over the table columns works reasonably well.  We could figure out how to expand the chart width to make the most of the available space, but currently they use a fixed width.

### DIFF
--- a/src/modes/top10/Top10.svelte
+++ b/src/modes/top10/Top10.svelte
@@ -271,10 +271,6 @@
     padding-top: 1em;
   }
 
-  td,
-  th {
-  }
-
   .table > table {
     border-spacing: 1em 0.5em;
     width: 100%;
--- a/src/modes/top10/Top10.svelte
+++ b/src/modes/top10/Top10.svelte
@@ -292,13 +292,13 @@
   }
 
   table .name {
-    width: 20em;
+    width: 10em;
     text-align: left;
     white-space: nowrap;
   }
 
   table .table-pop-column {
-    width: 5em;
+    width: 7em;
     text-align: right;
   }
 
@@ -308,17 +308,16 @@
   }
 
   table .main-indicator {
-    width: 30em;
+    width: 40em;
     text-align: right;
   }
 
   table .other-indicator {
-    width: 30em;
+    width: 40em;
     text-align: right;
   }
 
   table .add-column-container {
-    width: 30em;
     text-align: left;
   }
 
--- a/src/modes/top10/Top10.svelte
+++ b/src/modes/top10/Top10.svelte
@@ -273,11 +273,10 @@
 
   td,
   th {
-    border: 0;
   }
 
   .table > table {
-    border-collapse: collapse;
+    border-spacing: 1em 0.5em;
     width: 100%;
     overflow: unset;
   }
@@ -285,6 +284,42 @@
   table {
     padding: 1em;
     font-size: 1.2em;
+  }
+
+  table .rank {
+    width: 1em;
+    text-align: right;
+  }
+
+  table .name {
+    width: 20em;
+    text-align: left;
+    white-space: nowrap;
+  }
+
+  table .table-pop-column {
+    width: 5em;
+    text-align: right;
+  }
+
+  table th {
+    text-align: right;
+    white-space: nowrap;
+  }
+
+  table .main-indicator {
+    width: 30em;
+    text-align: right;
+  }
+
+  table .other-indicator {
+    width: 30em;
+    text-align: right;
+  }
+
+  table .add-column-container {
+    width: 30em;
+    text-align: left;
   }
 
   .go-to-map-pin {
@@ -343,8 +378,8 @@
 
   .remove-column {
     position: absolute;
-    right: 0.2em;
-    top: 0.2em;
+    right: 0;
+    top: -2em;
     font-size: 0.7rem;
   }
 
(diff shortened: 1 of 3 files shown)


dlaliberte:
Wide window with no additional indicator: 

image

dlaliberte:
Wide window with additional indicator:

image

Note the 'Cases' and 'Deaths' columns will get wider columns because the labels are longer.  This is a fortunate accident, but it is not terribly robust.

src/modes/top10/Top10SortHint.svelte
```diff
@@ -8,14 +8,15 @@
...
 <style>
   .root {
-    display: flex;
     cursor: pointer;
-    text-align: center;
-    justify-content: center;
     position: relative;
   }
 
+  .hint-wrapper {
```
sgratzl, line 15: please fix the warning posted by github action

src/modes/top10/Top10.svelte
```diff
@@ -287,6 +286,41 @@
...
+    width: 7em;
+    text-align: right;
+  }
+
+  table th {
+    text-align: right;
+    white-space: nowrap;
+  }
+
+  table .main-indicator {
```
sgratzl, line 310: why is the `table` selector in front of the classname selector needed?

sgratzl (commented):
see css selector comments, regarding UI not sure whether we want to right align the column title of e.g., `COVID-Like Symptoms`

sgratzl:
re UIKit adaptations: see also https://github.com/cmu-delphi/www-covidcast/pull/616 which uses the "native" UIKit table styles

src/modes/top10/Top10.svelte
dlaliberte: Generic class names like 'rank' and 'name' needed more qualification.  It's not needed for table-pop-column, but that could just be called 'population'.  (I thought 'pop' might be related to popper at first :))

I was also confused that .table is not used on a table but a div that contains the table.

I tend to make selectors more specific than needed, having experienced accidental collisions in the past, and adding more qualifiers also helps to make the meaning clear to readers. 

It is generally better to avoid use of tag names, e.g. table and th, so they might be replaced, but that's not so likely given table layout.

But I am open to your suggestions.

src/modes/top10/Top10SortHint.svelte
dlaliberte: done

src/modes/top10/Top10.svelte
sgratzl: re collisions: Svelte is automatically scoping each CSS class / selector. Thus, I'm usually not worried about CSS collisions.

dlaliberte:
> see css selector comments, regarding UI not sure whether we want to right align the column title of e.g., `COVID-Like Symptoms`

I found the previous layout very confusing regarding the alignment of the indicator column header and I think this layout is cleaner and more clear.  Either way, it might be good to also have another one or two headers above the value(s) to the left of the chart.

krivard:
It looks like this fix is limited to shuffling some of the extra space to between the pop and chart columns, which yields the same (though less severe) gestalt problem.

dlaliberte:
> re UIKit adaptations: see also #616 which uses the "native" UIKit table styles

We should apply your changes first.  Then I'll update mine again, assuming we still want to, given UI feedback.

dlaliberte:
> It looks like this fix is limited to shuffling some of the extra space to between the pop and chart columns, which yields the same (though less severe) gestalt problem.

I think Sam's changes in #616 are probably better regarding the column spacing.   The lines between rows probably help as well.    But there are several other changes here to consider as well as the shuffling of extra space:

- right-alignment of population numbers, and column header. 
- right-alignment of the indicator title.
- right-alignment of the rank number (leftmost column)
- left-alignment of the location name, and header.
- no white-space wrapping of the location name .
- move the "-" deletion icon for the extra indicator so it doesn't overlap the indicator title.

Letsgohyde:
Hey Guys, few recommendations for the default state list: 

1. Left align the "Name" header
image

2. Update % header, we can edit copy for this for each data type
image

3. Center "Add indicator" dropdown within far right <div> and center middle with chart to provide breathing room and visual separation from the chart col.
image

Overall reqs: 
image

let me know if I can provide more details :)

dlaliberte:
> 2. Update % header, we can edit copy for this for each data type

It looks like your "Percentage of visits" header is a variation of the indicator name, e.g. "COVID-related Doctor Visits".   It's not always a percentage. e.g. Bar visits.   And for the number of cases and number of deaths, there are two numbers, the raw number and the cumulative number. 

But it looks like you right-align the label, e.g. "Percentage of visits" on the number, with the date range left-aligned on the chart.  Is that intentional, or coincidental?

All the column headers (except the rank) are clickable, to sort the column first in descending order, then ascending.

PR #668, opened 2020-12-02
strip SWPA specific maps/code

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary

a while ago #381 was created in preparation of creating an Allegheny county specific map view focusing on this county and its neighborhoods. However, it seems that this code chunk will never be in production. Thus, I deleted it to cleanup the code.


tildechris (approved):
Negative PRs are the best.  Thanks for cleaning up.  LGTM.

PR #669, opened 2020-12-02
strip direction logic

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary
another dead code removing. In previous versions the API provided direction information (increasing, decreasing, steady). However, that has been removed a while ago since it was unstable and also removed from the map. 

The map has evolved since then making the old code outdated


dlaliberte:
I have no idea what the instability was previously, but I would think the direction indicator would be more useful if it was based on a moving average, so it doesn't switch direction so frequently.   But if this feature were reimplemented, would you do it much differently than it was previously?

sgratzl:
the code I striped are merely the last pieces which were left, even if we would have the direction attribute tomorrow, we would need several changes to get it working again. Moreover, we have the changes in the git history so it is not lost but just cleans up the current code.

dlaliberte (approved):

PR #671, opened 2020-12-02
simplify url parameters

### LINKED ISSUE #651: Support simple query strings
Right now, it appears that I have to use: 

https://covidcast.cmu.edu/?sensor=doctor-visits-smoothed_adj_cli&level=county&date=20201120&signalType=value&encoding=color&mode=export

if I were to provide somebody a link, and want them to land on the "Export Data" tab.  A lot of this info in the query string seems unnecessary for this tab, and conceptually it seems like all that should be needed is: 

https://covidcast.cmu.edu/?mode=export

This brings me to the "Export Data" tab but then it hangs for me (nothing loads).  FWIW, seems like it attempts to bring up this page: https://covidcast.cmu.edu/?mode=export&sensor=doctor-visits-smoothed_adj_cli&level=county&date=20201120&signalType=value&encoding=color

Can we support simple query strings like this one?  It would be great to do something similar for all the tabs: for each one, figure out the minimal amount of info that's needed to load this tab, and allow those query strings (rather than the full ones).

sgratzl:
closes #651 

**Prerequisites**:

- [x] Build is successful

### Summary

simplifies the tracked urls to only the needed one for the active move and also removes default values. This simplifies the export mode to e.g., http://localhost:8080/?mode=export


tildechris (approved):
LGTM.  This seems like a small improvement, but I really like how much simpler the initial experience looks with having few parameters.

PR #676, opened 2020-12-03
small map changes

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary

various changes related to the map
 * set a min/maximum zoom level
 * change automatic focus on region logic: only when the user uses the search bar it will focus on the selected region, in all other cases the zoom level of the map won't change automatically anymore.


tildechris (approved):
LGTM.  I didn't notice a big difference in how far you could zoom out, but the level to which you can zoom in looks like an improvement.  I also like the automatic focus changes.

Letsgohyde:
LGTM for the zoom-in limit

Also, not seeing the zoom-out limit, thank for sharing @sgratzl

PR #678, opened 2020-12-09
feat: generate new tab menu

sgratzl:
with icons and style

**Prerequisites**:

- [x] Build is successful

### Summary

generates a breadcrumb and a better tab menu row similar to the figma design:

image

*open*: right now we have `/covidcast/?mode=export` with some work on both sides (here and www-main), we could also have: `/covidcast/export` instead. Do we want that?


tildechris:
Looks good, but let's flag this for UX review.  @Letsgohyde can you take a look?

tildechris (approved):
LGTM.

PR #688, opened 2020-12-18
feat: default ratio cases/deaths

### LINKED ISSUE #536: consider switching to showing 'ratios per 100,000' per default
we should consider showing the ratio version for cases/deaths per default.

**User story**

while the count version is the default one in most maps, it inherits the problem of showing a population density map. The ratio version is normalized thus you can more easily see what are the current (normalized) trends. 

e.g.

http://localhost:8080/?sensor=indicator-combination-confirmed_7dav_incidence_num&level=county&date=20200930&signalType=value&encoding=color&mode=timelapse&region=42003

image

http://localhost:8080/?sensor=indicator-combination-confirmed_7dav_incidence_num&level=county&date=20200930&signalType=value&encoding=color&mode=timelapse&region=42003&signalR=true

image

Both pictures tell a very different story. However, one might consider this change a breaking change, since users who will go to the map will see a very different result. 

**Additional context**

sgratzl:
closes #536 

**Prerequisites**:

- [x] Build is successful

### Summary

REQUIRES: the suggested changes in https://docs.google.com/document/d/1RLy4O-gtACVjLEVD_vxyPqvp9nWVhqEjoWAKi68RKpg/edit#

changes the default cases/deaths signal to be the ratio one: 

image

including side plots

### DIFF
--- a/src/components/EncodingOptions.svelte
+++ b/src/components/EncodingOptions.svelte
@@ -53,27 +53,27 @@
     <div aria-label="display counts or ratios per 100,000 people" class="uk-button-group">
       <button
         type="button"
-        aria-pressed={String(!$signalCasesOrDeathOptions.ratio)}
+        aria-pressed={String(!$signalCasesOrDeathOptions.incidence)}
         class="uk-button uk-button-default uk-button-small"
-        class:uk-active={!$signalCasesOrDeathOptions.ratio}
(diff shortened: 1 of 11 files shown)


tildechris (approved):
LGTM, thanks Sam.

PR #704, opened 2020-12-30
unique bundle names

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary

creates unique bundle names by using webpacks `contenthash` within the file name. However, this requires https://github.com/cmu-delphi/www-main/pull/204 such that www-main also respects the new names.

This will avoid that the website and covidcast are out of sync on a client. A new version of the website will request a different filename from the server, thus a old browser cache hit won't be possible. An old version of the website will also request an old version of covidcast so also from this direction no problem anymore

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,12 +14,16 @@
   },
   "main": "public/bundle.js",
   "files": [
-    "public"
+    "public/**.js",
+    "public/**.txt",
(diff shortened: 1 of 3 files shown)


tildechris (approved):
LGTM

PR #708, opened 2021-01-04
feat: smart default sensor for fb-survey results

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary

changes the default sensor to a fb-survey one when the initial mode is survey-results. Thus, it will improve the default selected one to be the latest one according to the covidcast metadata


tildechris:
Is there a good reason that we shouldn't default to the FB latest date all of the time?  I don't think most users would expect that the date context between the map and the survey would be the same.

sgratzl:
> Is there a good reason that we shouldn't default to the FB latest date all of the time? I don't think most users would expect that the date context between the map and the survey would be the same.

so far the default sensor for COVIDcast is the 'doctor visits' one which has a lag of a few days. If we would use 'doctor visits' as default sensor to show in the map but use the latest date of the fb survey (lag of one or two days), we initially would always show an empty map. 

The current logic is that if the user is starting of with the map, you will see the latest doctor visits signal. If you start with the survey results and then switch to the map you will start with the fb-survey mask one.

tildechris:
What I'm proposing is that have a logically distinct date sensor for the survey page separate from the other COVIDcast pages.  

We could use a completely separate parameter from the `date` parameter that we already use, but that would break any deep links, so that is a no-go.

Alternatively, we could always reset to the latest date when switching to the survey page.  In this case, I think this is OK, since in COVIDcast, we always have a context sensor and switching over to the survey page logically switches that context to just the survey signals.  Thus, if we keep the `date` context between pages, it's always between two different signals, which I think is a pretty strong argument not to persist the selected date in this instance.

sgratzl:
> Alternatively, we could always reset to the latest date when switching to the survey page. In this case, I think this is OK, since in COVIDcast, we always have a context sensor and switching over to the survey page logically switches that context to just the survey signals. Thus, if we keep the `date` context between pages, it's always between two different signals, which I think is a pretty strong argument not to persist the selected date in this instance.

I changed the current logic, that upon switching to the survey-results page, the default survey signal of the map = mask is set along with its latest date.

tildechris (approved):
LGTM, thanks Sam.

PR #711, opened 2021-01-07
Add range selection to Details panel main chart

dlaliberte:
closes 

**Prerequisites**:

- [ ] Build is successful

### Summary

Experimenting with range selection.


dlaliberte:
How it looks at this time:

image

dlaliberte:
I can now compute the difference in values and draw a rule with arrowheads that is colored base on whether the change is an increase or decrease.

image

I also discovered a UI problem that will be difficult to work around.  Once the interval is selected, mouseover within the interval no longer will show tooltips for any points inside the interval.  This is because the mark drawn for the interval is above everything, by design in Vega.  There is discussion about allowing an option to specify the z-index of marks or layers, which would be useful in other circumstances, but no implementation of this, other than that the order of layers makes some difference, though it does not impact the selection drawing.

dlaliberte:
Here is what I have as of today:

image

dlaliberte:
Starting fresh, adding only what I needed, I managed to figure out the weird complication with multiple locations.

image

To do this, I am filtering the data for showing the area under the line to only include the 0th location.  Doing this filtering also avoids the random highlighting of one location or the other, but it means we would *always* *only* show the tooltip for the 0th location.  At least it is consistent.

tildechris:
I like how the shaded regions work.  The horizontal lines and the percentage indicator/lines.  Is there some place outside the plot area where we could have the start, end, and % difference text?

An alternate method would be to always have the context date (i.e. the red line) be the anchor.  Then when you hover over other points it would show you the % diff instead of having to click and drag to create the selection.

The more I think about it, it feels a bit odd to have the context date separate from a range selection.

One last note, have you played with making the discrete observations use a bar style instead of an interpolated line?

dlaliberte:
> I like how the shaded regions work. The horizontal lines and the percentage indicator/lines. Is there some place outside the plot area where we could have the start, end, and % difference text?

Yes, we could take some of the large space available to display data about the selected range.  We might do this in a way that is similar to what is being done on the survey results page, for example.  This would allow us to show more data, and avoid collisions with other data and graphics within the chart area.  Annotations within charts are hard to do well because of this.

> An alternate method would be to always have the context date (i.e. the red line) be the anchor. Then when you hover over other points it would show you the % diff instead of having to click and drag to create the selection.

Yes, I like this idea, but there are a couple complications.  

- The current date (aka context date) is not (yet) directly changeable on this detail chart page.  It can be changed by selecting a different date with the date control at the top, or by clicking on a different date in any of the small multiples on the right.  There is no reason I know of that we could not also allow clicking on a date in the detail chart page too. 
- Hovering over another date would *always* show this range selection, in addition to the tooltip for the hovered point.  If we do this, the range selection visualization should probably be a bit more subtle or light-weight, and then it would make more sense to show the associated data for the range in a separate location.
- There would be no way to preserve the range selection independent of hovering, not unless we add some additional capability, so it would disappear as soon as you stop hovering in the chart.  A modified click, such as shift-click, could make the range selection stick, but when would it go away?
- Touch-only support should be considered.

> The more I think about it, it feels a bit odd to have the context date separate from a range selection.

Yes, it does seem extraneous.

> One last note, have you played with making the discrete observations use a bar style instead of an interpolated line?

Yes, I have tried that, but got a mess with multiple locations that I didn't understand.  Now that I can filter out the data for other locations, it should be doable.  

We could also use a ["stepped-line" mode](https://www.google.com/search?q=stepped+line+series+chart&newwindow=1&sxsrf=ALeKk03vLQVAl5fF5m24wq5Ts64hH4ILVw:1611169120997&tbm=isch&source=iu&ictx=1&fir=b_psSQn2b8xkiM%252CtTPV1g8cQZkvYM%252C_&vet=1&usg=AI4_-kQ-fbW1tvbhUBkofPvtvT5UD39tZQ&sa=X&ved=2ahUKEwipzoTZmKvuAhXDmeAKHTMAD2MQ9QF6BAgOEAE#imgrc=b_psSQn2b8xkiM), which would be like the tops of the bars, rather than smoothed or straight lines (two alternatives) between points.

tildechris:
To capture our discussion from last week:

* I'd like to see a bar+line combo chart, but limit it to when there is only one location.  If there is more than one location, the chart should show only the separate 7-day average lines.
* We could move the tool-tip to follow the y-intercept on the series line instead of where the mouse pointer is.  This seems to be best practice on other sites

The range selection mechanism I think then would be different based on if there are multiple series.  If there's one series, you compare over time, but with multiple series, you compare over series.

Based on these observations, what do you think about closing this issue and starting a new one to focus on the bar+line chart?

PR #732, opened 2021-01-21
Click to select range to currentDate

dlaliberte:
closes 

**Prerequisites**:

- [ ] Build is successful

### Summary

This change keeps the original date range selector, but adds a click to 'select' a range between the date at a data point and the "current date".  Click again to change it or double-click to clear it.


dlaliberte:
What it looks like this way:

image

Hovering any points works within this selected range as well, but will always show the tooltip for the 0th location (if there is more than one location being displayed).

Issue #679, opened 2020-12-10
Consider alternatives to line charts


dlaliberte:
The line charts used in the small multiples and Detail View are well known, but might not be the easiest to use for all users. Here (and in subsequent comments) are a few examples of some alternatives that we might consider.

**Colored vertical lines in the background, corresponding to each data value.**  

The line is drawn on top so you can see the correspondence with each value.

![image](https://user-images.githubusercontent.com/570125/101717310-0da9a500-3a6d-11eb-88a9-18e37244d505.png)

- This only works with one location per chart.  Multiple locations would have to be plotted in separate charts.  
- The color range for each indicator might be different, and it should correspond to the same range used in the map for that indicator.
- Showing the error band around the data point (see Covid-like symptoms) has no correspondence in the background color.
- Technical problem: there is no automatic way to set the width of each vertical line so that it fills the background.  The pixel width has to be computed from the total screen width divided by the number of values.

dlaliberte:
Points colored by the same values.  When the adjacent values are close enough, this looks like a relatively solid line.  This could be used to create a colored "sparkline".

![image](https://user-images.githubusercontent.com/570125/101721269-8ca2db80-3a75-11eb-8dbc-7b9541e1b549.png)

- It is difficult to see the weekly cycles without the line.
- Creating a line that varies continuously in color doesn't seem possible.  It may be possible with discrete colors (I'll try that next), but should the color of the line before each point be based on the point, or the line after, or something else?

dlaliberte:
**Colored bars**

This is kind of a hybrid of the colored background and colored points.


![image](https://user-images.githubusercontent.com/570125/101721828-d93ae680-3a76-11eb-98aa-18dd7d8360b4.png)

dlaliberte:
In all these examples, I used the same color scheme, and let the colors scale automatically based on the values in this particular time range.  So the "People Wearing Masks" color varies more than it would if it were scaled to the baseline value of 0.  So this is a similar issue as auto-scaling the vertical axis to show more of the variance of the data.

dlaliberte:
Here is a recent blog post about [alternatives to line charts for time series data](https://adroitinsight.com/6-alternative-visuals-to-line-charts-for-time-series-data/) which happens to use covid-19 cases as data.

krivard:
[...]

PR #737, opened 2021-01-28
Add 1D heatmap to small multiple charts.

dlaliberte:
closes 

**Prerequisites**:

- [ ] Build is successful

### Summary

This draft PR adds the 1D heatmap to the small multiple charts, one of the alternatives that was considered in #679.

Here is how it looks on the top-10 page, with narrow window so that the 'mobile' mode kicks in:

image

Since this is the top-10 list, it shows multiple locations rather than multiple signals.  The same layout would apply either way, however.

Note: I haven't removed the line chart, or removed the vertical axis, and the colors need to be made consistent with the map colors.   Especially important is managing consistency regarding different signals with different scales, to avoid misleading interpretations of same or different colors across signals.


dlaliberte:
Here is what the small multiples (all signals) looks like without the line chart and y axis:

image

dlaliberte:
I also changed the Detail view to draw bars instead of the line chart, and add a moving average.  Looks like this:

image

dlaliberte:
Here is what it looks like with the error range smoothed, and drawn as an area around the smoothed data line.  The overlap with the bars is rather confusing.

image

dlaliberte:
With much lighter and thinner bars, and showing circles for the data points again, which might otherwise be missed entirely.

image

PR #738, opened 2021-01-29
Compute offset for tooltip from placement

### LINKED ISSUE #739: tooltips flicker when moving mouse off the chart to the right, and back again.
**Actual Behavior:**

Tooltips for the small multiples (also in top-10 and region details) flicker and remain partly obscured after moving the mouse off to the right of the chart and back over the chart.  Also happens when moving vertically off chart near the right end of the chart, followed by moving back onto the chart.

**Expected behavior**

Avoid flickering of tooltips.  Ensure they end up visibly placed without multiple redraws.

**Screenshots**

3

dlaliberte:
closes #739
**Prerequisites**:

- [x] Build is successful

### Summary

This change reduces tooltip flickering and improves placement to offset either right or left based on what placement is used (either top-start or top-end, due to the 'flip' option).    

This change also, perhaps coincidentally, substantially reduces the flickering and cut-off issues.  There is still one redraw that is cut-off before it decides to redraw again in flipped mode.

### DIFF
--- a/src/components/tooltipUtils.js
+++ b/src/components/tooltipUtils.js
@@ -35,7 +35,8 @@ function getOrInitPopper() {
       {
         name: 'offset',
         options: {
-          offset: ({ placement }) => (placement.endsWith('start') ? [8, 8] : [-8, 8]), // [0, 8],
+          // Compute the offset from the placement.
+          offset: ({ placement }) => (placement.endsWith('start') ? [8, 8] : [-8, 8]),
         },
       },
       {
--- a/src/components/tooltipUtils.js
+++ b/src/components/tooltipUtils.js
@@ -35,7 +35,7 @@ function getOrInitPopper() {
       {
         name: 'offset',
         options: {
-          offset: [0, 8],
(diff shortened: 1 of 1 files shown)


sgratzl:
please add a screencast for the before and after behavior

dlaliberte:
Before screencast added to #739.  
Here is a screencast for the fixed behavior:

http://somup.com/crnVoPo7JX

sgratzl (approved):
thx for the screencasts

👍 

we could also think of setting the default placement to `top-end` instead of `top-start`, such that by default it will place it to the and only flip it to the right if needed. Since you side panel is to the right that might be useful

dlaliberte:
I tried changing the default to 'top-end', but then it *always* draws first on the right, and then immediately flips to the left.  This might be a bug in popper, or perhaps there is a way to change the timing of the initial drawing so that it delays drawing until it has figured out what the default is.  By leaving the default as 'top-start', at least it avoids the redraw unless it needs to.

PR #747, opened 2021-02-09
Add ChangeIndicator, for tooltips

dlaliberte:
closes 

**Prerequisites**:

- [x] Build is successful

### Summary

Compare data values where hovering with data at 'currentDate'.


dlaliberte:
Here is how it looks on the Details panel for the Cases data, which has multiple parts.

image

Do we want to show changes for all these parts of the data?

tildechris:
> Do we want to show changes for all these parts of the data?

IMHO, I think only having counts/pop is needed in the pop-over.  Thoughts?

PR #769, opened 2021-03-03
add per 100k unit

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary

based on our discussion from yesterday, I created a similar per 100 unit sign and applied it:

image

image

image

### DIFF
--- a/src/components/UIKitHint.svelte
+++ b/src/components/UIKitHint.svelte
@@ -7,7 +7,7 @@
 
 <style>
   .title {
-    color: initial;
+    color: inherit;
     width: 12px;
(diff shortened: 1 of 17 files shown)


tildechris (approved):
I per 100K looks beautiful on mobile; we'll have to tweak the desktop styles.

LGTM, only nit is that there's some commented dead code.

sgratzl:
> LGTM, only nit is that there's some commented dead code.

you are referring to the global trends. I prepared the code for some helper methods to compute trends/min/max on the whole time window and not just the 4 month window.

PR #771, opened 2021-03-03
apply latest styles to survey results

sgratzl:
closes 

**Prerequisites**:

- [ ] Build is successful


tildechris:
Since we're leaving out the y-axis label, should we always add a title to the time-series restating the indicator and units?  Also it looks like line height needs to be tweaked on desktop (looked fine on mobile):

image

sgratzl:
image

sgratzl:
Note: I also changed the indicator text from increasing/decreasing to up and down (less letters)

src/modes/mobile/RegionHexMap.svelte
```diff
@@ -79,6 +80,17 @@
   $: colorScale = sensor.createColorScale($stats, region.level);
 
   const maxColumn = state2TileCell.reduce((acc, v) => Math.max(acc, v.x), 0) + 1;
+
+  function isInvertedColor(v) {
+    const color = v && v.value != null ? colorScale(v.value) : MISSING_COLOR;
+    return hsl(color).l < 0.5;
+  }
+
+  function style(v) {
```
tildechris, line 89: Thanks for this refactor, this is much clearer than having the logic inline.

src/stores/params.js
```diff
@@ -211,6 +213,150 @@ export class DataFetcher {
...
+    return sensors.map((sensor) => this.fetch1Sensor1RegionNDates(sensor, region, timeFrame));
+  }
+
+  /**
+   * @param {(Sensor|SensorParam)[]} sensors
+   * @param {Region[]} regions
+   * @param {TimeFrame} timeFrame
+   * @return {Promise<EpiDataRow[]>[]}
+   */
+  fetchNSensorNRegionNDates(sensors, regions, timeFrame) {
```
tildechris, line 273: These fetch methods read pretty well.  If we were to migrate some of this logic into the API layer, how much of an improvement would that be in load time and reducing front-end complexity?

tildechris (approved):
LGTM.  Thanks for advocating for this change.  My only concern is that the results page feels a little more "data savvy", but I link having consistent conventions better and I think it's still approachable.

src/stores/params.js
sgratzl: the following features for the api will make things simpler
 * possibility to fetch multiple signal at a time (so far only possible for the same data source)
 * no more max rows = no batching needed => will be part of flask implementation
 * multiple geo types at once (there is an issue for that), e..g nation and state at the same time
 * min/mean/max per given region / date frame -> easier trend computation

PR #795, opened 2021-03-09
chart labels

### LINKED ISSUE #792: Time-series labeling: Add titles, X-axis descriptions and attribution.
Since #787 gives us the ability to download full charts at a time, we need to make sure each is labeled so that it can stand on it's own.  This means we should:

* [x] Define the y-axis unit in the chart title.
* [x] Add "Delphi Group, delphi.cmu.edu/covidcast" attribution as a small caption
* [ ] Define the x-axis unit more exactly.  Almost all of our charts are time-series, but we should define which date is being talked about.  For example, with cases and deaths, it's "Reported Date" (vs. the actual test or death date)
* [ ] The current spacing in desktop is too tight.  There shouldn't be more than 5 or 6 x-axis labels.
* [x] We should add the year to the x-axis.  If the time-range doesn't span new year's, the year should only appear once, if it does it should appear twice:

```
|*---------*---------*---------*---------*---------*---------*---------|
 9/30/20             10/30               11/30               12/30

|*---------*---------*---------*---------*-|-------*---------*---------|
 10/30/20            11/30               1/1/21              2/30
 ```

sgratzl:
based on #787 

closes #792 

**Prerequisites**:

- [x] Build is successful

### Summary

image

image


sgratzl:
image

sgratzl:
> Thanks, I can see how mobile wraps the title, and I think it looks good. For the survey results page, now that we're being more precise with our titles, should we remove the section titles to be a little less redundant? I also think we don't need the "Click on the chart to select a different date." text.

done

tildechris (approved):
LGTM.  This looks really nice, I'm super excited to get this feature out there.

PR #800, opened 2021-03-10
adapt export mode

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary

adapt the export mode to match the other styles. Not perfect but better what we have:

image


tildechris (approved):
LGTM.  One minor tweak:

image

When the user uses the search for a region box, can we select the "specific regions" radio button for them?  I was surprised to see all counties when I downloaded a signal and it was because I didn't also change the radio button.

PR #802, opened 2021-03-10
use per 100 unit

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary

switches to `per 100` [people] unit and uses the relative change for showing trends


sgratzl:
did we agree on a strategy regarding the unit in the indicator and geo table?

tildechris:
> did we agree on a strategy regarding the unit in the indicator and geo table?

I think this PR captures the consensus.  The % is only used for relative change and units should be expressed in either per 100 or per 100K.

tildechris (approved):
LGTM.

PR #833, opened 2021-03-17
feat: adapt landing text

sgratzl:
closes #828 

**Prerequisites**:

- [x] Unless it is a hotfix it should be merged against the `dev` branch
- [x] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [x] Build is successful
- [x] Code is cleaned up and formatted

### Summary

localhost_8080_(iPhone X)

image

PR #829, opened 2021-03-17
feat: rename pages

sgratzl:
**Prerequisites**:

- [x] Unless it is a hotfix it should be merged against the `dev` branch
- [x] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [x] Build is successful
- [x] Code is cleaned up and formatted

### Summary

changes the title of the pages according to the new "explore" scheme, see also counterpart in www-main soon

PR #834, opened 2021-03-17
unify google docs / meta data description

sgratzl:
based on #833 #829

**Prerequisites**:

- [x] Build is successful

### Summary

larger refactoring to unify the more and more diverging survey questions, classic map signal descriptions and v2.0 signal descriptions. 

visually it should look the same. noticeable changes:
 * the "wanted test" now shows as a unit `per 100 people out of who didn't get a test` (see slack thread)
 * access to the raw signals in the export data tab

Note: as long as all my suggestions aren't applied in the google docs this build will fail


src/App.svelte
```diff
@@ -22,11 +22,11 @@
   <div class="loader loading" />
 {:then value}
   <svelte:component this={value} />
-{:catch error}
+  <!-- {:catch error}
```
tildechris, line 25: Nit: Dead code?

tildechris (approved):
LGTM.

PR #853, opened 2021-03-23
show data anomaly annotations

### LINKED ISSUE #601: Consider supporting an explainer popup to provide additional details on exceptional phenomena
All of humanity's technology systems are learning how to handle a global pandemic at the same time, and we should expect a commensurate number of errors and weird edge cases. If we could provide users with greater context for interpreting these scenarios, it would both increase trust in our tool and help elevate the public's understanding of the scientific process.

**User story**

The health department of Wyoming has been including probable (but untested) cases in their case count for months. When they decide to switch to only including tested confirmed cases, the apparent case incidence plummets to huge negative numbers. This shows up in the map as Wyoming being a sole buttercup in a field of dark red. A visual cue lets the user know that yes, this is the correct display, and it is indeed weird. When clicked, it brings up a card that explains the additional context, and links to the relevant announcement on the Wyoming DoH website.

Screenshot lifted from @tildechris's outages report:

image

**Additional context:**

In addition to the bulk updates and negative incidence events that happen from time to time, this feature could also be used to explain things like:
* Why some counties are grouped together (they share public health resources and don't report separately)
* Why the daily count in NYC counties is sometimes fractional

sgratzl:
closes #601 

current state: 
image

sub tasks
 * [ ] define api -> which URL should be used?
 * [x] load annotations
 * [x] design annotations
 * [x] add annotations to history line chart
 * [x] add annotations to general indicator overview
 * [x] add annotations to tables?
 * [ ] add annotations to maps?

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,10 +3,13 @@
 
 [context.branch-deploy.environment]
   COVIDCAST_ENDPOINT_URL = "https://api.covidcast.cmu.edu/epidata/api.php"
+  COVIDCAST_ANNOTATION_DRAFTS = "true"
 
 [context.deploy-preview.environment]
   COVIDCAST_ENDPOINT_URL = "https://api.covidcast.cmu.edu/epidata/api.php"
+  COVIDCAST_ANNOTATION_DRAFTS = "true"
 
 [context.staging.environment]
   COVIDCAST_ENDPOINT_URL = "https://api-staging.delphi.cmu.edu/epidata/api.php"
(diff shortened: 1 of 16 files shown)


sgratzl:
history line annotations:

image

Letsgohyde:
This is looking great, few design changes: 

1. Can we reorder the alert header copy: Icon, "Low spiky values" [date range]
2. Increase contrast: Darker shade of orange for the text and icon. #D46B08

Thanks @sgratzl

sgratzl:
tables:

image

image

tildechris (approved):
LGTM! Great job, this is a surprisingly concise change.

PR #863, opened 2021-03-25
advanced workflows

sgratzl:
### Summary

adds additional github actions to handle the release


.github/workflows/update_gdocs_data.yml
```diff
@@ -0,0 +1,35 @@
+name: Update Google Docs Data
```
tildechris, line 1: @krivard FYI - not just releases but updating the code from the docs update is automatable via this action.

sgratzl: once it is deployed, I will add docu/link to the Google Doc with instructions

tildechris (approved):
LGTM.  We live in the future.

PR #865, opened 2021-03-26
Correlation metrics

tildechris:
**Prerequisites**:

- [x] Build is successful

### Summary

Add a utility method that generates overall r^2, and then over a +/- 28 day period max r^2, lag at max and a series such that the r^2 vs. lag can be plotted.


tildechris:
The regression library does add heft to the bundled js.  This logic should be moved to the server when possible.

sgratzl:
> The regression library does add heft to the bundled js. This logic should be moved to the server when possible.

do you mean bundle size wise? as far as I can see it is actually a quite small library

package.json
```diff
@@ -100,7 +100,7 @@
     "mapbox-gl": "^1.13.0",
     "marked": "^2.0.1",
     "mini-css-extract-plugin": "^1.3.9",
-    "node-fetch": "^2.6.1",
```
sgratzl, line 103: why did you remove this dependency?

src/data/testData.js
```diff
@@ -0,0 +1,3151 @@
+export let deathsTestData = [
```
sgratzl, line 1: what about moving this test data to some directory like `data/__tests__/testData.js` to make it clear it is for tests only?

src/data/utils.js
```diff
@@ -38,3 +39,83 @@ export function parseAPITime(v) {
...
+/**
+ * Generates R^2 metrics for lags between -28 and 28 days.
+ *
+ * For lags between 0 and 28 lag b backwards with respect to a.  For -28 to -1 lag a with
+ * respect to b.
+ *
+ * For each lag, the input is a window of length(a)-28, such that the number of values at
+ * each lag is the same.
+ *
+ * @param {Number[]} a
```
sgratzl, line 60: prefer `number[]` instead of `Number[]`

src/data/utils.js
```diff
@@ -38,3 +39,83 @@ export function parseAPITime(v) {
...
+ * @param {Number[]} b
```
sgratzl, line 61: prefer also adding a `@returns {type}` hint

src/data/utils.js
```diff
@@ -38,3 +39,83 @@ export function parseAPITime(v) {
...
+ */
+function generateLags(a, b) {
+  let lag = 28;
```
sgratzl, line 64: prefer `const` instead of `let` when possible

src/data/utils.js
```diff
@@ -38,3 +39,83 @@ export function parseAPITime(v) {
...
+    return [x, b[i]];
+  });
+}
+
+/**
+ * Compute 28-day correlation metrics for a response variable given an explanatory variable.
+ *
+ * @param {import('./fetchData').EpiDataRow[]} response
+ * @param {import('./fetchData').EpiDataRow[]} explanatory
+ * @param {CorrelationMetric} metric
```
sgratzl, line 99: isn't that `@returns`?

src/data/utils.js
```diff
@@ -38,3 +39,83 @@ export function parseAPITime(v) {
...
+ */
+export function generateCorrelationMetrics(response, explanatory) {
+  let response_values = response.map((row) => {
```
sgratzl, line 102: prefer inline arrow function such as `(row) => row.value`, easier to read for such simple functions

src/data/utils.js
```diff
@@ -38,3 +39,83 @@ export function parseAPITime(v) {
...
+  });
+  let explanatory_values = explanatory.map((row) => {
+    return row.value;
+  });
+
+  let model = linear(zip(response_values, explanatory_values));
+  let lags = generateLags(response_values, explanatory_values);
+  let max = lags.reduce((acc, i) => {
+    return i.r2 > acc.r2 ? i : acc;
+  });
```
sgratzl, line 113: missing reduce initial argument?

sgratzl (commented):
in general I prefer using `const` over `let` when possible. This gives also the compiler more hints about how to optimize the code

package.json
tildechris: I was using it initially in the unit test to use the API, then reverted my changes when I extracted test data.  I didn't notice that it was used elsewhere.  Will revert.

src/data/testData.js
tildechris: Good suggestion.  I didn't see an existing spot.

src/data/utils.spec.js
```diff
@@ -22,3 +23,18 @@ describe('parseAPITime', () => {
...
+
+describe('correlationMetrics', () => {
+  test('Deaths vs. CLI correlations should equal what was published in the blog post', async () => {
+    let expected_metrics = {
+      r2At0: 0.43,
+      lagAtMaxR2: 20,
+      r2AtMaxR2: 0.81,
+    };
+    let actual_metrics = generateCorrelationMetrics(deathsTestData, cliTestData);
+    expect(actual_metrics.r2At0).toEqual(expected_metrics.r2At0);
```
sgratzl, line 35: an alternative method to defining the values directly is to use snapshot testing: https://jestjs.io/docs/snapshot-testing

src/data/utils.js
```diff
@@ -38,3 +39,87 @@ export function parseAPITime(v) {
...
+    lags.push({ lag: -1 * i, r2: model.r2 });
+  }
+  return lags;
+}
+
+/**
+ * Do a pair-wise combination of the elements from a and b arrays.
+ * @param {number[]} a
+ * @param {number[]} b
+ * @returns {number[]}
```
sgratzl, line 94: isn't it `[number, number][]` a tuple of two numbers?

tildechris:
> > The regression library does add heft to the bundled js. This logic should be moved to the server when possible.
> 
> do you mean bundle size wise? as far as I can see it is actually a quite small library

I thought I saw a number of transitive dependencies, but you're right it's pretty lightweight.  Still would make sense to do server side, however.

src/data/utils.spec.js
tildechris: That's a pretty neat feature.  Here's however, I wanted to be a little looser about the assertion, specifically in lags array that is returned.  Currently it's not sorted with respect to lag number, so if the ordering changes in the future, I didn't want to break the test.

src/data/utils.js
tildechris: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) says that "If no initialValue is supplied, the first element in the array will be used as the initial accumulator value and skipped as currentValue".

sgratzl (approved):
👍

PR #877, opened 2021-03-30
Add correlation metrics to the table

tildechris:
Adds metrics to the correlation table on the indicator details page.


src/modes/mobile/IndicatorCorrelation.svelte
```diff
@@ -22,7 +32,34 @@
...
+      .filter((d) => d.key !== sensor.key)
+      .map((sensor) => {
+        return {
+          ...sensor,
+          metrics: buildMetrics(sensor),
+        };
+      });
+  }
+
+  $: otherSensors = buildTableData();
```
tildechris, line 62: @sgratzl I don't have a full mental model of how Svelte handle reactive updates.  This renders fine on initial load, but then doesn't update based on the user's suggestion.

sgratzl: the code in the function won't be tracked. So you have to pass all the tracked/reactive variables, in this case `sensor, region, date`

tildechris: Ah, I see.  It's only reactive with respect to the declarations that have changed. Makes sense.  Thanks.

src/data/utils.js
```diff
@@ -107,17 +128,19 @@ function zip(a, b) {
  * @returns {CorrelationMetric}
  */
 export function generateCorrelationMetrics(response, explanatory) {
-  const response_values = response.map((row) => row.value);
-  const explanatory_values = explanatory.map((row) => row.value);
+  const zippedEpiData = zipEpiDataRow(response, explanatory);
```
sgratzl, line 131: what happens if there is no overlap or no given data?, e.g. if we have no data for a selected region?

src/modes/mobile/IndicatorCorrelation.svelte
```diff
@@ -52,6 +92,15 @@
             {sensor.name}
           </a>
         </td>
+        {#await sensor.metrics}
+          <td><p>...</p></td>
+          <td><p>...</p></td>
+          <td><p>...</p></td>
+        {:then m}
+          <td><p>{m.r2At0}</p></td>
```
sgratzl, line 100: prefer using some formatter from `./src/format.js`, e.g. `formatValue`

src/data/utils.js
tildechris: Good question.  If there's not at least 28*2 shared elements we should return a message saying there wasn't enough data.

src/modes/mobile/IndicatorCorrelation.svelte
tildechris: Will take a look.

src/data/utils.js
```diff
@@ -107,17 +139,24 @@ function zip(a, b) {
...
+  const explanatoryValues = zippedEpiData.map((row) => row[1]);
 
-  const model = linear(zip(response_values, explanatory_values));
-  const lags = generateLags(response_values, explanatory_values);
+  const lags = generateLags(responseValues, explanatoryValues);
   const max = lags.reduce((acc, i) => {
     return i.r2 > acc.r2 ? i : acc;
   });
 
+  const lagAtZero = lags.filter((l) => l.lag == 0)[0];
```
sgratzl, line 156: prefer `.find` instead of `.filter(...)[0]`

sgratzl (approved):
just a small comment but besides that looks good

src/data/utils.js
```diff
@@ -98,6 +103,33 @@ function zip(a, b) {
     return [x, b[i]];
   });
 }
+/**
+ * Do a pair-wise intersection of EpiDataRow by date.
```
dlaliberte, line 107: Maybe this is more straightforward code, but I am doing a similar thing for one pair of sensors, without depending on the ordering of the time_value, with the following function:

```
  let sensorListData = [
    { sensor: primary, data: primaryData },
    { sensor: secondary, data: secondaryData },
  ];

  $: sensorCorrelationData = mergeSensors();
  // for each time_value, merge data values across sensors.
  function mergeSensors() {
    const sensorDateMap = {};
    const sensorKeysMap = {}; // map from sensor key to sensor.
    sensorListData.map((sensorAndData) => {
      const { sensor, data } = sensorAndData;
      data.forEach((row) => {
        const time_value_key = String(row.time_value);
        if (!sensorDateMap[time_value_key]) {
          sensorDateMap[time_value_key] = { ...row };
        }
        const sensorKey = sensor.key;
        sensorKeysMap[sensorKey] = sensor;
        sensorDateMap[time_value_key][sensorKey] = row.value;
      });
    });
    return Object.values(sensorDateMap);
  }
```

I did it this way because I was merging rows for all the sensors, with slightly more complex code.

tildechris: If you think this operation is generalizable, I can export intersectEpiDataRow and have it return two new `EpiDataRow[]` using my previous `Map` based algorithm to be tolerant of order.

Perhaps additionally a new method that returns a `Map<Date, EpiDataRow>`.

sgratzl: see also https://github.com/cmu-delphi/www-covidcast/blob/sgratzl/typescript/src/data/utils.ts (TS version for the types)

tildechris: I'll create another PR for additional utility methods.

PR #893, opened 2021-04-05
Annotation Overview Table

### LINKED ISSUE #879: create "Annotations Overview Page"
**User story**

Review all annotations by time.  Being able to see all annotations gives one a good idea of the challenges of data reporting over time.

sgratzl:
closes #879

**Prerequisites**:

- [x] Build is successful

### Summary

adds a new page: `/data-anomalies` which shows the plain list of data anomalies similar to the Excel sheet in a nice way:

image

Moreover, I refactored the sort table logic for a simpler reuse

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -35,6 +35,9 @@
                 <li data-mode="indicator-status" class="nav-dropdown-child">
                   <a href="/indicator-status">Indicator Status Overview</a>
                 </li>
+                <li data-mode="data-anomalies" class="nav-dropdown-child">
+                  <a href="/data-anomalies">Data Anomalies</a>
+                </li>
                 <li data-mode="export" class="nav-dropdown-child"><a href="/export">Export Data</a></li>
                 <li class="uk-nav-divider"></li>
                 <li data-mode="lab" class="nav-dropdown-child"><a href="/lab">Lab</a></li>
--- a/src/modes/data-anomalies/AnnotationTable.svelte
+++ b/src/modes/data-anomalies/AnnotationTable.svelte
@@ -23,7 +23,7 @@
     };
   }
 
-  const sort = new SortHelper('source', false, 'dateRange');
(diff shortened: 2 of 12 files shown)


krivard:
@sgratzl This is good, though I think I want to add a more informative label than Indicator Combination to those cases/deaths entries. Should I do that through a new column in the spreadsheet?

@Letsgohyde I also sense a blurring of intended audience here and am open to ideas on how to resolve. Nontechnical users will definitely care why cases suddenly appeared to increase by 300%, and technical users will care what API queries are affected (so they'll need the exact signal names)... but I don't want to suggest to nontechnical users that understanding the signal names is essential/fundamental to interpreting the data anomaly listed.

sgratzl:
> @sgratzl This is good, though I think I want to add a more informative label than Indicator Combination to those cases/deaths entries. Should I do that through a new column in the spreadsheet?

for the export tab I explicitly split the cases/death signals from the remaining indicator-combination ones, we can do a similar thing here. 

image

krivard:
> explicitly split the cases/death signals

perfect! let's do it

sgratzl:
> @Letsgohyde I also sense a blurring of intended audience here and am open to ideas on how to resolve. Nontechnical users will definitely care why cases suddenly appeared to increase by 300%, and technical users will care what API queries are affected (so they'll need the exact signal names)... but I don't want to suggest to nontechnical users that understanding the signal names is essential/fundamental to interpreting the data anomaly listed.

I assume for non-technical users they will encounter those anomalies just as annotations/warning in the other pages. I thought this page just to be more for data exporters that wanna have an overview.

sgratzl:
as discussed I split the "public health reports" data source in "COVID Cases" and "COVID Deaths"

chinandrew (approved):
:shipit: :shipit: :shipit:

PR #903, opened 2021-04-15
COVIDcast dashboard

sgratzl:
exploratory building block approach for COVIDcast:

image

image

features:
 * numerous widgets
 * flexible layout
 * persisted in the URL
 * linking and brushing between widgets
 * configure parameters (indicator, region, date) that is picked up my default by the widgets


sgratzl:
new features
 * choropleth map value highlight (next to legend)
 * highlight hhs region when hovering over state

sgratzl:
added option to configure default params = url params:

image

krivard (approved):
xlnt

PR #930, opened 2021-05-05
migrate to covidcast api 2.0

sgratzl:
WIP

**Prerequisites**:

- [x] Unless it is a hotfix it should be merged against the `dev` branch
- [x] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [x] Build is successful
- [x] Code is cleaned up and formatted

### Summary

migrates to the new covidcast api custom endpoints

* [x] correlation
* [ ] trend
* [ ] meta
* [x] csv

PR #931, opened 2021-05-07
backfill profiling viz

sgratzl:
based on #930 

adds a new section to the indicator status overview page about the backfill profile of the selected indicator:

image

https://deploy-preview-931--cmu-delphi-covidcast.netlify.app/indicator-status/#change_b

### DIFF
--- a/src/modes/indicator-status/IndicatorBackfill.svelte
+++ b/src/modes/indicator-status/IndicatorBackfill.svelte
@@ -172,7 +172,7 @@
   </div>
 </div>
 <AboutSection details className="uk-margin-small-top uk-margin-small-bottom">
-  <h3 slot="header" class="mobile-h3">DayOfWeek Distributions</h3>
+  <h3 slot="header" class="mobile-h3">Day of Week Distributions</h3>
   <div class="desc">
     In order to highlight day of week changes within reported values of the selected indicator the following charts are
     stratified by day of week. The horizontal axis shows the lag while the vertical axis shows the fraction of value or


krivard:
I'm going to show this at leads tomorrow afternoon for more feedback

krivard:
@RoniRos do you know yet what you want in place of "completeness" here?

RoniRos:
Very nice!   I do have some suggestions:

1.  In the "Region" box, COVIDcast usually fills in in faint ink "Search for state or country".  Should we do it here, too? 
2. When I switch between the "coverage details" tab and the "Backfill Profile" tab, I'd like to maintain the signal and location.
3. "Reported Date" and "Issue Date" aren't quite the right terms.  "Issue" is a Delphi-internal term which is not known to many others (maybe only those who have worked with our Epidata before).  "Reported Date" is ambiguous, because it could be the day being reported on, or the day the reporting is done.  How about "Reference Date" and "Reporting Date"?  I welcome a discussion of these terms - they will be with us for a long time, and it's important to get them right!   Whatever we choose, a concise definition should be easily accessed from this page (e.g. by a link or tooltip from these phrases).  For example:
     * Reference Date: the date of the activity being measured (e.g. "Date of Service" for doctor visits, or "specimen collection date" for lab test results).
     * Reporting Date: the date on which the specific value was reported to Delphi.
       (We will likely run into multiple versions of Reporting Dates, for example the date a value was available at one of our sources vs. the date at which it was incorporated into Delphi.  I suggest for now we stick with my definition above, i.e. reported to Delphi). 
4. Instead of having to introduce the concept of "Anchor" and define "Completeness" relative to it, how about we replace it with "Signal value as fraction of its value ___ days after Reference Date", with ___ being an input field defaulting to 60.4. The "Date" pull down menu should be called something like "Show by:"
5. The "Color" menu could be renamed "Tracked value", and could contain:
   * "As fraction of value at lag=60" or "As fraction of value 60 days after Reference Date".
   * "Day-to-day relative change"
   * I do not understand what "sample size" is and what it measures.
6. In the day-to-day relative change mode, the color legend doesn't look right -- it should be bi-color, no? 
 Also should be labeled appropriately.  Also, what is the blue line?
7."Lag" could be explained by changing the Y Axis title from "Lag" to "Lag (Report Date - Reference Date)"
8. The indicator names should be more descriptive.
9. Generally, I think "signal" might be better than "value" in most contexts here.  "Value" is just so generic -- every number is a value.

sgratzl:
> * In the "Region" box, COVIDcast usually fills in in faint ink "Search for state or country".  Should we do it here, too?
> * When I switch between the "coverage details" tab and the "Backfill Profile" tab, I'd like to maintain the signal and location.
> 
> * "Reported Date" and "Issue Date" aren't quite the right terms.  "Issue" is a Delphi-internal term which is not known to many others (maybe only those who have worked with our Epidata before).  "Reported Date" is ambiguous, because it could be the day being reported on, or the day the reporting is done.  How about "Reference Date" and "Reporting Date"?  I welcome a discussion of these terms - they will be with us for a long time, and it's important to get them right!   Whatever we choose, a concise definition should be easily accessed from this page (e.g. by a link or tooltip from these phrases).  For example:
>   
>   * Reference Date: the date of the activity being measured (e.g. "Date of Service" for doctor visits, or "specimen collection date" for lab test results).
>   * Reporting Date: the date on which the specific value was reported to Delphi.
>     (We will likely run into multiple versions of Reporting Dates, for example the date a value was available at one of our sources vs. the date at which it was incorporated into Delphi.  I suggest for now we stick with my definition above, i.e. reported to Delphi).
> 
> * Instead of having to introduce the concept of "Anchor" and define "Completeness" relative to it, how about we replace it with "Signal value as fraction of its value ___ days after Reference Date", with ___ being an input field defaulting to 60.4. The "Date" pull down menu should be called something like "Show by:"
> 
> * The "Color" menu could be renamed "Tracked value", and could contain:
>   
>   * "As fraction of value at lag=60" or "As fraction of value 60 days after Reference Date".
>   * "Day-to-day relative change"

done all above

>   * I do not understand what "sample size" is and what it measures.
> 
> * In the day-to-day relative change mode, the color legend doesn't look right -- it should be bi-color, no?

image

in this example it is divergent color scale but the change of 150 is overwhelming the visual legend (-0.45 ... 150)

>   Also should be labeled appropriately.  Also, what is the blue line?

removed for the day to day change

>   7."Lag" could be explained by changing the Y Axis title from "Lag" to "Lag (Report Date - Reference Date)"

done

> 
> * The indicator names should be more descriptive.

see issus in delphi-epidata

> 
> * Generally, I think "signal" might be better than "value" in most contexts here.  "Value" is just so generic -- every number is a value.

RoniRos:
> the change of 150 is overwhelming the visual legend (-0.45 ... 150)

I don't understand.  I can see beige streaks in the image, but I don't see any beige in the color legend.
Does the color legend go from -0.99999 to +\infty?   Shouldn't we use a log scale to stretch the negative values?  We can still label the map with non-log-scaled values.

Also:

* I can specify a lag>60 (e.g. 90), and it happily uses this value in various places in the display, but it looks like the analysis is actually capped at lag=60.

* Is there a way to clarify that in the phrase "Lag  (Reporting Date - Reference Date)", the hyphen refers to numerical subtraction rather than a range?   Perhaps a rephrasing?  E.g. "Lag (in days, from Reference Date to Reporting Date)"

> When I switch between the "coverage details" tab and the "Backfill Profile" tab, I'd like to maintain the signal and location.

The location is now indeed remembered in the "Backfill Profile" tab, but the signal does not appear to be preserved, nor to transfer back-and-forth between the two tabs.

krivard:
> I can specify a lag>60 (e.g. 90) ... but it looks like the analysis is actually capped at lag=60.

That's a limitation in the database -- all the claims signals submit dense (not diff-based) issues, and currently have the following lag caps:

* doctor-visits: 70
* hospital-admissions: 70
* chng: 60

RoniRos:
> In theory you could pull these numbers out of metadata.max_lag, but that is a descriptive statistic and not an SLA statistic, so it's been polluted with long lags from repairs and bugfixes.

I don't fully follow the considerations here.  But in general, if I can tell from the display that the selected anchor is not being used, then maybe the program can detect that too, and let the user know that it's not going to respect their requested anchor.  For example, when I ask for an anchor of 90, the heatmap still goes only to 60 or 65.

sgratzl:
> I don't understand. I can see beige streaks in the image, but I don't see any beige in the color legend.
> Does the color legend go from -0.99999 to +\infty? Shouldn't we use a log scale to stretch the negative values? We can still label the map with non-log-scaled values.

the color scale in this example goes from -0.45 ... 0 ... 150. Thus if you have like 150 pixels to show this linear color scale it would mean less than 1 pixel for the are between -0.45 and 0 and almost 150 pixel for the area of 0 ... 150. However, we could/should clamp the scale, because a value of +150 = the value is +15000% to the previous one, seems a far outlier. A possible maximum value could be 2 = +200%, or 5.

sgratzl:
> "Typical max lag" would be a good candidate to add to a facts table for each source+signal, along with #539 and perhaps #544.

I don't get the relation to the referenced issues, can you clarify?

sgratzl:
> the heatmap still goes only to 60 or 65.

the anchor lag so far can be interpreted as: the last lag/issue that has a lag <= anchor lag

For example, consider the following two dates and their available lags:
2021-01-01 has issues with lags: 4, 6, 10, 12, 34
2021-01-02 has issues with lags: 4, 5, 11, 33

let's assume anchor lag =12. 
For 2021-01-01 it will pick 12 and for 2021-01-02 it will pick 11, since both both represent the issues that would be valid at a lag of 12. 

The blue anchor line in the heatmap shows which anchor / lag has been used for this specific date. In your example, you requested that you wanna use an anchor of at most 90. However, the data only has issues with a maximal lag of 60 to 65 (see comment form Katie). Thus, those were picked as the anchor lag.

sgratzl:
> The location is now indeed remembered in the "Backfill Profile" tab, but the signal does not appear to be preserved, nor to transfer back-and-forth between the two tabs.

now I understand what you meant, should be fixed.

sgratzl:
re descriptions (e.g. to clarify lag): https://docs.google.com/document/d/1x6Yfr4pZFexJc3vxei37CI4nCGKh04-yin6Ea0CWpYE/edit

RoniRos:
I like your idea of clamping the scale, maybe to 2 or 3?  If the value changes day to day by a factor of more than 3, it probably doesn't matter by how much it changes.

However I think the scale should be symmetric in the log domain: the equivalent of a bump up by a factor of 2 is a bump down by a factor of 2.  If you clamp the bump up at 3, let's clamp the bump down at 1/3.   For legibility, the tickmarks should probably be labeled in the original domain, namely something like (*1/3+ ,*1/2,  -30%,  -20%,  -10%, NC, +10%, +20%,  +30%,  *2, *3+).  I realize these are not exactly symmetric, and that the notation can be improved, but hopefully the gist is clear:  we want to see small changes, which we usually think of as +/-X%, and we want to know when the changes become so big that they change the order of magnitude.

sgratzl:
> However I think the scale should be symmetric in the log domain: the equivalent of a bump up by a factor of 2 is a bump down by a factor of 2. If you clamp the bump up at 3, let's clamp the bump down at 1/3. For legibility, the tickmarks should probably be labeled in the original domain, namely something like (*1/3+ ,*1/2, -30%, -20%, -10%, NC, +10%, +20%, +30%, *2, *3+). I realize these are not exactly symmetric, and that the notation can be improved, but hopefully the gist is clear: we want to see small changes, which we usually think of as +/-X%, and we want to know when the changes become so big that they change the order of magnitude.

re scale: the charting library has some limitations but I tried to achieve what you described:

image

RoniRos:
> > However I think the scale should be symmetric in the log domain: the equivalent of a bump up by a factor of 2 is a bump down by a factor of 2. If you clamp the bump up at 3, let's clamp the bump down at 1/3. For legibility, the tickmarks should probably be labeled in the original domain, namely something like (*1/3+ ,*1/2, -30%, -20%, -10%, NC, +10%, +20%, +30%, *2, *3+). I realize these are not exactly symmetric, and that the notation can be improved, but hopefully the gist is clear: we want to see small changes, which we usually think of as +/-X%, and we want to know when the changes become so big that they change the order of magnitude.
> 
> re scale: the charting library has some limitations but I tried to achieve what you described:
> 
> image

Looks good!

* The "NC" (no change, alternative notation welcome) should be easily distinguished from small changes, because it usually corresponds to no updates, which is important to know.  Ideally, it would be represented by white.  Is this already the case?  Namely, is "NC" represented by a distinct white, easily distinguishable from e.g. +/- 3% ?

* Re the majority color (greyish brown?), am I correct in assuming that these are mostly "down" corrections?  Or is it the color of "NC"?  Ideally, it should be easy to see when there are "down" and when there are "up" corrections.  If it's not clear, we may want to clamp at (1/2, 2) and stretch the colors a bit.

* is there a good way to indicate that the edges of the color legend are clamped, namely that "*3" is actually "times 3 or more", and "*1/3" is actually "divided by 3 or more"?  Also 1/3 might be easier to mentally parse if rendered as proper fractions (\frac{1}{3}).

sgratzl:
I switched to this color scale: https://github.com/d3/d3-scale-chromatic#interpolateRdBu and added the white = 0 as suggested. Those color scales are designed that the values around 0 have some neutral color, since small changes around 0 might be the result of numeric errors. 

image

RoniRos:
> I switched to this color scale: https://github.com/d3/d3-scale-chromatic#interpolateRdBu and added the white = 0 as suggested. Those color scales are designed that the values around 0 have some neutral color, since small changes around 0 might be the result of numeric errors.
> 
> image

I like this!  Thank you!

RoniRos:
* In "Fraction of value at lag 60" mode, report dates that didn't have issues (e.g. 5/9 in New York Change) show as white.  Shouldn't they show the column to their left (values of the previous report date), based on our definition of 'as-of', which is cumulative?
* "Fraction of value at lag __" can easily exceed 1.0, and in fact it often does.  Right now you are clamping the color scale at 1.0 .  I think we should clamp it at higher than 1, preferably based on the data (say at the 95% quantile?), otherwise at some compromise value (e.g. 1.25).  Also, it might make sense to use a color scale that shows qualitatively which values are above 1.0 vs. which are below.
*

sgratzl:
>     * "Fraction of value at lag __" can easily exceed 1.0, and in fact it often does.  Right now you are clamping the color scale at 1.0 .  I think we should clamp it at higher than 1, preferably based on the data (say at the 95% quantile?), otherwise at some compromise value (e.g. 1.25).  Also, it might make sense to use a color scale that shows qualitatively which values are above 1.0 vs. which are below.

e.g. 

image

sgratzl:
>     * In "Fraction of value at lag 60" mode, report dates that didn't have issues (e.g. 5/9 in New York Change) show as white.  Shouldn't they show the column to their left (values of the previous report date), based on our definition of 'as-of', which is cumulative?

I would keep them in a white/missing color, since e.g., 

image

gives you a nice idea, when no issues were triggered, e.g. because of some outtake in the pipeline)

krivard:
@RoniRos this has been great feedback! I'd like to release these next week. We can always patch changes to the color maps later; do you have any dealbreakers that must be fixed before the first release?

sgratzl:
> @RoniRos this has been great feedback! I'd like to release these next week. We can always patch changes to the color maps later; do you have any dealbreakers that must be fixed before the first release?

ping

RoniRos:
Sorry, I thought we communicated on this in another channels (comments on a googledoc?)  Anyway, I agree we should release ASAP!  My understanding was that the only release-blocking potential issue was the signal naming.  Is it finalized (or at least finalized enough)?  Is it waiting for me?

krivard (approved):

sgratzl:
I prefer changing one thing at a time, so let's merge and alter it later

Issue #928, opened 2021-05-03
dashboard: support absolute counts for cases/death values


sgratzl:
COVIDcast 2.0 is using the per 100k version only. However, for the general dashboard it would be good to have an option to show also the absolute cases/death values in addition to the per100k versions, so that the total variation space includes:

* absolute / per 100k
* cumulative / incidence
* smoothed / raw
* cases / deaths

PR #936, opened 2021-05-12
cumulative support for history line charts

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary

add support to show the raw cumulative signal version for cases/deaths. part of #928 

image

note: the dark green portion is the amount that has been added for this day

### DIFF
--- a/src/blocks/HistoryLineChart.svelte
+++ b/src/blocks/HistoryLineChart.svelte
@@ -13,6 +13,7 @@
     MULTI_COLORS,
     genDateHighlight,
     genAnnotationLayer,
+    generateCumulatedBarSpec,
   } from '../specs/lineSpec';
   import { toTimeValue } from '../data/utils';
   import Toggle from '../components/Toggle.svelte';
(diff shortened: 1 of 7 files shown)


krivard:
While "cumulated" is a legit word, it tends to be reserved for far more technical contexts than we need here. [Use "cumulative" instead](https://books.google.com/ngrams/graph?content=cumulated%2Ccumulative).

I don't understand what "raw: 16.6" means in the caption, since the cumulative data is also raw (unsmoothed). do you mean "incident"? "new" would also work there

sgratzl:
so far: `raw` = unsmoothed and `cumulative` = unsmoothed cumulative. I don't know whether `new` or `incident` would be more clear in this case

sgratzl:
> While "cumulated" is a legit word, it tends to be reserved for far more technical contexts than we need here. [Use "cumulative" instead](https://books.google.com/ngrams/graph?content=cumulated%2Ccumulative).

done

PR #944, opened 2021-05-20
fix: county regions are not found

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary

fixes a bug where the state of a region couldn't be resolved

original: 
> On the [dev site](https://dev--cmu-delphi-covidcast.netlify.app/indicator/?date=20210513&sensor=indicator-combination-confirmed_7dav_incidence_prop&region=54007) (but not in the production site), going to "Explore an Indicator", selecting "Cases", then double-clicking on WV, then clicking on one of the counties, creates confusion:  COVIDcast claims that the 1,792,000 population of WV is actually the population of that county, and also that every county hovered over is called "Calhoun County".


krivard (approved):
👍  nice work

PR #945, opened 2021-05-20
improve Vega performance

sgratzl:
based on #943 

**Prerequisites**:

- [x] Build is successful

### Summary

improves the rendering performance of Vega especially for the county map

1. refactored the vega.svelte such that less updates are triggered
2. set `precision = 0` such that the geometric data won't be adaptive sampled. it might end up a few artifacts but since we mostly have straight lines we shouldn't see much differences.


krivard:
more artifacts than we were hoping, particularly around state borders:

image


the technologies are for sure completely different, but maybe there's something we can take from https://github.com/cmu-delphi/covidcast/issues/542?

sgratzl:
current dev version: 

image

after creating a "highres" version of county and states for detail views: I don't think it can better with the base shape files that we have (https://github.com/cmu-delphi/covidcast/tree/main/Python-packages/covidcast-py/covidcast/shapefiles)

image

krivard:
How much faster is it? if it's faster by a lot, we can accept the weird nonmatching outlines and just flag finding better shape files as future work. if it's not that much faster, might as well leave it how it is.

sgratzl:
> How much faster is it? if it's faster by a lot, we can accept the weird nonmatching outlines and just flag finding better shape files as future work. if it's not that much faster, might as well leave it how it is.

the first version of the previous comment was showing the current dev version without this improvements/PR applied: 

dev version:

https://user-images.githubusercontent.com/4129778/119189725-6a60b100-ba4a-11eb-804d-2328ca2d490f.png

your posted image:

image

I cannot see any differences between the current dev version of this image and the one you have posted. Can you elaborate how this PR changed the view?

krivard (approved):
👍  sorry about that -- I definitely forgot to compare to existing behavior...

PR #987, opened 2021-07-20
export tab flat list of signals

### LINKED ISSUE #984: Export Data variant selection is confusing
**Actual Behavior:**

Variant selection does not clearly present the available options. The user is presented with two select boxes: the first uses the prose name for one of the base signals available from this source, and the second uses the API name for the variants available for the selected base signal.

API names are used nowhere else on the website, so this is likely to confuse users for whom the Export Data page is their only access to the data. Additionally, some signals in the meta sheet use prose names which are specific to the variant, making the text insufficiently general to use as a label for a collection of variants.

The fix will likely require work across both www-covidcast and the meta sheet.

1. Go to https://delphi.cmu.edu/covidcast/export/
2. Select any of the cases/deaths signals
3. Observe SELECT SIGNAL

**Expected behavior**

The names in the first select box should be general enough to cover all variants.

Individual variants should be legible to users unfamiliar with the API.

The intended variant-specificity of names and descriptions in the meta sheet should be treated uniformly across all signals, and documented in the Note on the column header in the sheet.

A few implementation possibilities:

1. Make the signal name column suitable for all variants. Make the Short Description specific to the variation and suitable for use in the second select box.
2. Add a new column to the signals meta sheet that stores a prose name to use for all variants of a base signal. Make the signal names of all variations specific to the variation and suitable for use in the second select box. 
3. Use the "source subdivision" meta sheet functionality to store the prose name to use for all variants of a base signal. Make the signal names of all variations specific to the variation and suitable for use in the second select box. 
4. Add a new sheet for collections of variants, and use this instead of the current "base signal" approach

also open to others. I do notice the category chips and the description text for each variant shows up below the select boxes, but that doesn't help defray the initial confusion, only helps soothe it after the user has clicked around experimentally, and only if that part of the interface is visible while they do so.

**Screenshots**
image

**Rating Scale (1 is minor, 7 is severe):** 4 -- confusing; ultimately nonblocking, but will reduce our conversion rate

**Context**

- Browser:Chrome

sgratzl:
closes #984 

**Prerequisites**:

- [x] Build is successful

### Summary

changes back to a flat list of signals: 

image

will be useful once each signal has an unique name

### DIFF
--- a/src/modes/exportdata/ExportData.svelte
+++ b/src/modes/exportdata/ExportData.svelte
@@ -17,24 +17,12 @@
 
   let sourceValue = null;
   $: source = sourceValue ? $metaDataManager.metaSources.find((d) => d.source === sourceValue) : null;
-  $: sensors = source ? source.sensors.filter((d) => d.meta.signal_basename === d.signal) : [];
-  let sensorBaseValue = null;
-  $: sensorBase = sensorBaseValue ? $metaDataManager.getSensor(sensorBaseValue) : null;
-
-  $: {
-    if (source && !sensors.find((d) => d.key === sensorBaseValue)) {
-      sensorBaseValue = sensors[0].key;
-    }
-  }
-  $: sensorVariants =
-    sensorBase && source
-      ? source.sensors.filter((d) => d.meta.signal_basename === sensorBase.meta.signal_basename)
-      : [];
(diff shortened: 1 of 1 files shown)


krivard (approved):
👍

PR #992, opened 2021-07-30
fix: msa mixed signal

### LINKED ISSUE #991: Survey results dashboard showing sawtooth pattern when MSAs are selected
**Actual Behavior:**

When an MSA is selected from the region dropdown, the plots on the survey results dashboard show a sawtooth pattern. 

1. go to https://delphi.cmu.edu/covidcast/survey-results/?date=20210727&region=38300
2. Observe sawtooth pattern in time series plots

It's likely this is because the interface is trying to plot a context curve (national, given the queries I see in the browser's network log) but then dumping both sets of points in a single curve. No context curve is described in the legend.

**Expected behavior**

Either of these would be fine:
* Don't permit the user to select MSAs
* Separate the points into two curves

**Screenshots**

image

**Rating Scale (1 is minor, 7 is severe):** 6, impossible to interpret MSA plots

**Context**

- Browser: Chrome

**Additional Context**

I've verified that the API response for each of the queries run by the browser includes only one point for each date for each time series requested, so this is not a repeat of the multiple is-latest-issue flags problem.

[Slack thread with report ](https://delphi-org.slack.com/archives/C0130CSQRN3/p1627498514071500)and approval from Alex to nix MSAs if we need to

sgratzl:
closes #991 

**Prerequisites**:

- [x] Build is successful

### Summary

image

### DIFF
--- a/src/blocks/HistoryLineChart.svelte
+++ b/src/blocks/HistoryLineChart.svelte
@@ -91,10 +91,10 @@
   $: timeFrame = showAllDates ? sensor.timeFrame : date.windowTimeFrame;
 
   /**
-   * @param {import('../../stores/params').SensorParam} sensor
-   * @param {import('../../stores/params').RegionParam} region
-   * @param {import('../../stores/params').DateParam} date
-   * @param {import('../../stores/params').TimeFrame} timeFrame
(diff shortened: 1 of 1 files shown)


krivard (approved):
🎉

PR #1012, opened 2021-09-20
refactor google docs and remove duplicates

### LINKED ISSUE #1010: Remove redundant information from signals doc
When I was messing around with #1009 I noticed that it was still displaying "per 100 people" even though that text never appeared in the source. It turned out this text was coming from `descriptions.raw.txt` and `descriptions.generated.json`, which both get refreshed by `npm start`. These files are read by `constants.ts`, and from there, used everywhere. To prevent these files from forcing outdated information into the web visualizations, we should remove all redundant fields from each listing.

[Link to signals description doc](https://docs.google.com/document/d/1llv6xh8jMlmVv7WpyDSv4VgUpFAZQ6QjFUeRxxOinmk/edit#)

sgratzl:
closes #1010 

**Prerequisites**:

- [ ] Build is successful

### Summary

changes the code to avoid duplicates from the google docs and reduces them to only the necessary elements. 

For this I was working on two copies of the docs such that the generation works:  https://docs.google.com/document/d/1IcpYAG721QFetdp1Lx2EeXGNITbB37ELJtmsKzk6pnk/edit# and https://docs.google.com/document/d/1xh3XdBI43sYmoS6q1mPMcvKAbxfwX8aRqs8zaxCAnjg/edit#

However, atm. two signals are not delivered by the metadata: 

image

see also https://github.com/cmu-delphi/delphi-epidata/issues/717 and https://github.com/cmu-delphi/delphi-epidata/pull/716 

Note: the build will fail since this PR doesn't match the current google docs anymore.


krivard (approved):
👍

PR #1013, opened 2021-09-20
fix: avoid percentage of per 100

### LINKED ISSUE #1008: Percentages are double-labeled with both % and "per 100"
**Actual Behavior:**

Percentage signals display as "X% per 100", which is difficult to interpret.

1. go to https://delphi.cmu.edu/covidcast/indicator/?date=20210907&sensor=fb-survey-smoothed_wcli
2. observe e.g. "4 month worst value of 1.69% per 100 on Mon, Aug 30th 2021."
3. Line chart is okay
4. Legend on line chart says "1.67% per 100"

Additionally,
1. https://delphi.cmu.edu/covidcast/indicator/?date=20210907
2. Line chart says this is per 100 people, but this signal is in fact per 100 outpatient visits, not per population at all

**Expected behavior**

* X%
* X per 100
* X percent

Probably best to avoid "per 100 people" unless we're sure it's a percentage of people and not something else like encounters. We could consider something like

* X per 100 [items]

if we felt like specifying `items` for each signal in the advanced meta, but that seems like work.

**Rating Scale (1 is minor, 7 is severe):** 3 - confusing, but we've gone this far without noticing it so...

**Additional context**

We received a question about this from an MD user in the IHS.

sgratzl:
closes #1008 based on #1012

**Prerequisites**:

- [x] Build is successful

### Summary

before: 

image

after:

image

notes
 the build fails since there are some changed in the Google doc before this PR that haven't been incorporated.


krivard:
Ah, we do want % for changes though. "Better by -0.20" --> "Better by -20%", since otherwise it sounds like we're saying 5.2 = 6.3 - 0.20

image

& same in the"Increasing" chip and the "Change last 7 days" columns

sgratzl:
> Ah, we do want % for changes though. "Better by -0.20" --> "Better by -20%", since otherwise it sounds like we're saying 5.2 = 6.3 - 0.20

fixed

krivard (approved):
👍

PR #1021, opened 2021-09-25
show stderr and auto scale in CTIS dashboard

### LINKED ISSUE #1019: Consider changing display of charts in surveys dash to make small changes more visible
Several of the surveys signals change on a very small scale, and recent trends are muted or invisible when using a 0-anchored y-axis. We want to consider the following changes:

* switch the default display to have "Re-scale Y axis" turned on
* make uncertainty explicit using error bars / shaded region showing stderr

0-anchored Y axes are accepted best practice for bar charts, but for line charts, recommendations vary. In the case of the survey signals where small changes in public opinion or behavior may have larger effects on public health, it makes sense to zoom in to make these small changes more visible by default.

Adding in the stderr will help prevent overamplification of trends/noise.

**Additional context**

Before we go through with this change, we need to review the stderr computation of non-CLI/ILI survey signals to make sure they are correctly accounting for survey weights.

### LINKED ISSUE #1023: Check computation of y-value bounds
**Actual Behavior:**

> In the attached screenshot, the y-range isn't being properly computed. Perhaps it has something to do with missing values?
image

sgratzl:
closes #1019, closes #1023

**Prerequisites**:

- [x] Build is successful

### Summary

image

### DIFF
--- a/src/blocks/HistoryLineChart.svelte
+++ b/src/blocks/HistoryLineChart.svelte
@@ -280,7 +280,7 @@
     return spec;
   }
 
-  export let zoom = false;
+  export let zero = true;
   export let stderr = false;
   let singleRaw = false;
   let singleCumulative = false;
@@ -294,7 +294,7 @@
   $: spec = injectRanges(
     genSpec(sensor, region, date, timeFrame, {
       height,
-      zero: !zoom,
+      zero,
       raw,
       isMobile: $isMobileDevice,
       singleRegionOnly,
@@ -350,7 +350,7 @@
 />
 
 <div class="buttons">
-  <Toggle bind:checked={zoom}>Rescale Y-axis</Toggle>
+  <Toggle bind:checked={zero}>Include 0 in Y-axis</Toggle>
   {#if sensor.rawValue != null && !($isMobileDevice && showAllDates)}
     <Toggle bind:checked={singleRaw}>Raw Data</Toggle>
     {#if raw && sensor.rawCumulativeValue != null}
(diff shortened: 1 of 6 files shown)


sgratzl:
@ryantibs since you are testing the PR, please also comment here, on any changes that you would like to see.

ryantibs:
Thanks!  I'd like to see:

- Survey dashboard plot titles restored to what they are in the current live version.
- Rescale Y-axis" toggle to say "Include 0 in Y-axis", and then set it to be "off" by default. (Same effective behavior, just different name---I'm open to tweak for the new name as well.)
- Y-range to be properly computed always, there are currently bugs (as in #1023).

It's possible I'd like to see more.  I am working through several thoughts on related things now as I start looking at the dashboard and main viz site more.  @sgratzl: Would you like me to also comment on those on this PR?  Or somewhere else?

ryantibs:
@sgratzl Ok so I think I'll save the other major things I'd like to see changed for a different conversation (I'll write it up in a google doc as we discussed).

Here's one more small thing though for the CTIS dashboard.  Where it says: _The 7-day average of "Went to Work or School" on 09/24/2021 is based on 180,890 samples with a standard error of ±0.1. It was published on 09/26/2021._  

I'd like to remove the part that reports the standard error here; it's already visible in the figure and it seems like a strange choice to display it there in this sentence anyway, where there is no reference to the value of the signal itself.

sgratzl:
2. and 5. should be implemented. I extracted the rest to new issues, since they are not directly related to the issue #1019 which is the base of this PR

krivard:
Whitespace -- "Include 0 in Y-axis" label is closer to the toggle for Raw Data than to its own toggle

image

sgratzl:
re styling:
image

krivard (approved):
👍

ryantibs:
@krivard I'm looking at the CTIS dashboard now and notice that this still hasn't been ported over to the live version.  Is that because we are waiting to launch all the new changes to the dashboards at once?  (Sam is still working on things for the main COVIDcast dashboard, but that's nothing to do with the CTIS dashboard AFAIK?)

PR #1035, opened 2021-09-28
Restore old survey question names

### LINKED ISSUE #1022: Automatic Signal Names differ from previously configured ones
**Actual Behavior:**

see comment from different issue:
> The plot titles look off from the live version. For example, I see the word "(Weighted)" in the plot title now in a bunch of places (attached screenshot is an example), where I don't see this in the live version. Is this a bug or just something that will be automatically fixed in the live version?
image

reason: 

which the change to the sheet based metadata all signals got new names and suffixes. While in the previous version the names were explicitly configured, #1012 changed it to use the meta data directly resulting in a name change. 

possible options: 
 * in the remaining google doc, explicitly override the name for signals
 * leave it, since it is more descriptive anyhow
 * change signal name in meta data

sgratzl:
closes #1022

**Prerequisites**:

- [x] Build is successful

### Summary

image

### DIFF
--- a/src/modes/survey-results/SurveyQuestion.svelte
+++ b/src/modes/survey-results/SurveyQuestion.svelte
@@ -33,11 +33,19 @@
 
   export let anchor = null;
 
-  $: sensor = new SensorParam($metaDataManager.getSensor(question), $metaDataManager);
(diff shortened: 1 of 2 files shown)


sgratzl:
wontfix

Issue #1032, opened 2021-09-28
Restore CTIS Dashboard Signal Names


sgratzl:
**Actual Behavior:**

due to the use of the new meta the signal names have changed. 

> Survey dashboard plot titles restored to what they are in the current live version.

sgratzl:
Duplicate of #1022

PR #1038, opened 2021-09-28
add "Metro Area" suffix

### LINKED ISSUE #1031: MSAs being wrongly labeled as cities
MSAs are not labeled as such.  For MSAs identified by a single city, this results in strongly misleading labels.  For example, in the classic view, if you search for "Pittsburgh, PA" and display, e.g. cases, you get a title "COVID cases in Pittsburgh, PA", but it's really the Pittsburgh MSA (2.36M people, a superset of Allegheny County), NOT the city of Pittsburgh (~300K people, a subset of Allegheny County).  Similarly for all other cities and signals.

The problem stems from the fact that the MSAs get their names from the city or cities on which they are centered.  To fix this, the label should include the "MSA" phrase, e.g. 
  * In places requiring brevity: "Pittsburgh PA MSA".
  * In places allowing a longer label (like in the title):  "Pittsburgh, PA Metropolitan Statistical Area"

The "MSA" qualifier should also be included in the completion options in the location search box.

sgratzl:
closes #1031 
based on #1032 

**Prerequisites**:

- [x] Build is successful

### Summary

image

### DIFF
--- a/src/data/regions.ts
+++ b/src/data/regions.ts
@@ -96,7 +96,7 @@ export const nationInfo: RegionInfo = {
   population: stateInfo.reduce((acc, v) => acc + (v.population ?? 0), 0),
 };
 
-export const msaInfo = parseCSV(msaRaw, 'msa');
+export const msaInfo = parseCSV(msaRaw, 'msa', (entry) => `${entry.name!} Metro Area`);
 export const countyInfo = parseCSV(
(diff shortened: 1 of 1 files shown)


krivard (approved):
Approving "Metro Area" append; signals rename revert pending feedback on #1022

PR #1060, opened 2021-10-02
Improve Indicator Table

### LINKED ISSUE #1055: Indicator Table Improvements
- [x] rename COVID-19 Indicators to COVID-19 Indicators Table
 - [x] remove type filter dropdown
 - [x] The breakdown into Public Behavior, Early Indicators, Cases and Testing, Late Indicators (recall: third one is a new category), can go as vertical text spanning the table rows.
 - [x] source name can go in parentheses next to the signal name.
 - [x] We can also try to make the sparkline smaller (shrink to take up less vertical space, but preserve the aspect ratio: I like the current aspect ratio, when the table is at full width).
 - [x] Column rename: CHANGE LAST 7 DAYS → RELATIVE CHANGE LAST 7 DAYS
 - [x] When the value is NA, the relative change should be NA (not 0.00%).
 - [x] Arrows/colors should be disabled since the trend calls are.
 - [x] Column justification is weird to me: can't tell if it's right or left justified but I don't know why the value is right next to the sparkline, and yet there is a big space between the relative change and the value
 - [ ]

sgratzl:
closes #1055 

**Prerequisites**:

- [x] Build is successful

### Summary

image

### DIFF
--- a/src/data/api.ts
+++ b/src/data/api.ts
@@ -257,7 +257,7 @@ export interface EpiDataMetaStatsInfo {
   stdev: number;
 }
 
-export type SignalCategory = 'public' | 'early' | 'late' | 'other';
+export type SignalCategory = 'public' | 'early' | 'late' | 'other' | 'cases_testing';
 export type SignalFormat = 'raw' | 'percent' | 'fraction' | 'per100k' | 'count';
 export type SignalHighValuesAre = 'good' | 'bad' | 'neutral';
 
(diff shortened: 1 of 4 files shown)


krivard:
Minor nit: can we make N/A and unit-less values align with the right edge of values instead of the right edge of units?
image

-->
image

sgratzl:
how about that:
image

krivard (approved):
⭐

PR #1063, opened 2021-10-05
Improve GeoTable

### LINKED ISSUE #1053: Geo Table Improvements
- [ ] remove rows with NA values and indicate at the bottom that XX locations have been omitted because of missing values

sgratzl:
closes #1053 
based on #1060 since both are the same basic table

**Prerequisites**:

- [x] Build is successful

### Summary

image

### DIFF
--- a/src/data/api.ts
+++ b/src/data/api.ts
@@ -257,7 +257,7 @@ export interface EpiDataMetaStatsInfo {
   stdev: number;
 }
 
-export type SignalCategory = 'public' | 'early' | 'late' | 'other';
+export type SignalCategory = 'public' | 'early' | 'late' | 'other' | 'cases_testing';
 export type SignalFormat = 'raw' | 'percent' | 'fraction' | 'per100k' | 'count';
 export type SignalHighValuesAre = 'good' | 'bad' | 'neutral';
 
--- a/src/data/sensor.ts
+++ b/src/data/sensor.ts
@@ -10,11 +10,15 @@ import type { TimeFrame } from './TimeFrame';
 export const sensorTypes: { id: SignalCategory; label: string }[] = [
   {
     id: 'public',
-    label: 'Public’s Behavior',
+    label: 'Public Behavior',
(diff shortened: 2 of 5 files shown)


krivard:
Why does it say 5 locations are hidden for missing values when the hex map only shows one state with NA?

[People Wearing Masks on Sept 30](https://deploy-preview-1063--cmu-delphi-covidcast.netlify.app/indicator/?utm_source=github&utm_campaign=bot_dp&date=20210930&sensor=fb-survey-smoothed_wwearing_mask_7d)

image
image

sgratzl:
looking at https://delphi.cmu.edu/covidcast/indicator/?date=20210930&sensor=fb-survey-smoothed_wwearing_mask_7d

the missing ones are the territories which are not part of the hex m ap

krivard:
That's reasonable then; can we list the hidden regions in a tooltip? Maybe just the first 5 "and {n-5} others" to avoid sparse signals causing problems on state pages.

krivard:
@ryantibs this is what I mean when I say prevents comparisons: 

Data is available for Maui County on September 30 but not on October 1. If you have September 30 selected, you can see how Maui County compares to the other Hawaii counties in the table:

image


But if you have October 1 selected, Maui County gets dropped, even though there are dates in the sparkline with available data:

image

& if you want to find out what days do have Maui County data, you have to go to the Maui County page. If you want to find out if there are any days that have data for all counties in Hawaii, you have to click through the calendar one day at a time. This seems suboptimal.

Is this the desired behavior, or did you just mean to hide regions where everything is NA, not just the selected day?

ryantibs:
@krivard Good question/catch.  Yes I meant we should remove rows where **all data** is NA.  And I guess we could do interpret this to mean all data in the sparkline.

There could still be some "surprises" where locations magically appear as you move back or forward in time (all data is gone for ~1 month, but then a single data point appears), but I think that will be rare.

PR #1065, opened 2021-10-05
set default date to today()-1

### LINKED ISSUE #1056: Improve Current Date Selection
I suggest we use current day - 1.  By my count this means (depending on what time of day the site is visited?) the following should be available: 
 * All public health reporting signals: cases, hospitalizations, deaths (just confirming: this would be true with HHS hospitalizations signal)?
 * All CTIS signals: vaccine, masks, two CLI signals (one CLI signals is further lagged by 1 day but that is being fixed now)

By my count, here is how the other signals behave: 
 * Google signal will be 3 days lagged
 * Doctor visits will be 4 days lagged
 * Quidel signal will be 5 days lagged
 * CAN signal will be 3 days lagged

sgratzl:
closes #1056

**Prerequisites**:

- [x] Build is successful

### Summary

change the default date to today - 1

image


krivard:
This is still draft so you may already know this but something in the logic is not quite right; it's saying there isn't data for october 4 when there clearly is:

image

krivard:
Yep, meta data only gets updated nightly, and we won't be able to switch to realtime metadata updates until we fix a problem causing random meta update runs to take 4 hours.

The true max_time_value can be up to 24 hours off for signals that update daily, and up to 7 days off for signals that update weekly (currently just bar & restaurant visits, but some of the upcoming flu signals will only get weekly updates as well).

What are our options? I see these:
* leave the banner as-is, and just eat the fact that it is sometimes comically wrong
* drop the banner
* to compute whether to show the banner, look in metadata first. if it says data is available, believe it. if it says data is not available yet, check the actual signal response.

any others?

ryantibs:
@sgratzl I'm trying to revisit this to help shepherd v3 along to being finished.  Can you explain to me what the purpose of the banner is---is it only to display a warning if the data isn't available (what's being shown in orange in the above screenshots)?  

Anyway regardless, unless the banner is for a purpose that I really don't understand, it seems like Katie's suggest solution in bullet point 3 is fine.

sgratzl:
the goal of this banner was to give users an information why they see only N/A for the whole page, e.g. in a normal case it could look like:

image

> to compute whether to show the banner, look in metadata first. if it says data is available, believe it. if it says data is not available yet, check the actual signal response.

done

ryantibs:
Ok looks good thanks.

ryantibs (commented):
It looks good to me, I just want to double-check that the feature I suggested "for each indicator in the table, report the value and relative change for the most recent date available in the table instead of NA, if they're not available at the current time" is being implemented on some other PR.

(Because the table still shows NAs to me.)

sgratzl:
> It looks good to me, I just want to double-check that the feature I suggested "for each indicator in the table, report the value and relative change for the most recent date available in the table instead of NA, if they're not available at the current time" is being implemented on some other PR.

yes that will come later, this PR is a prerequisite for that one. 

please approve the PR via GitHub, otherwise I cannot merge it

ryantibs (approved):
Yep I was going to approve it (doing so now) just checking I understood what was going on with my last comment.  Thanks!

PR #1068, opened 2021-10-17
update signals

### LINKED ISSUE #1042: Remove SafeGraph signals
> I don't learn anything from the raw signals, because they're so occluded by weekend/weekday effects.  And also because mobility patterns really haven't been all that interesting thus far.   I don't have any evidence that anybody has used any of these so far.  I'd like to just remove them from the dashboard entirely.

### LINKED ISSUE #1043: Replace Hospitalization Signal
> I would like to include this signal everywhere in the dashboard.  To be clear, I'm talking about: confirmed_admissions_covid_1d_prop_7dav.  And I would be in favor of removing the claims-based hospitalization signal everywhere from the dashboard (to avoid any ensuing confusion about why we have two hospitalization signals).
>
> I'm aware that the HHS hospitalization signal is only state-level.  I think that is fine for now.  I have some thoughts about what to do at the county-level (everything would be a week old at best, based on the facility-level data set) but I'll shelve them for now.

### LINKED ISSUE #1048: Add COVID Act Now (PCR) signal

### LINKED ISSUE #1052: use Cases signal as new default
I would like COVID cases to be the default, not doctor visits.

### LINKED ISSUE #1054: Explore a location Improvements
- [ ] add Cases, Hospitalizations and Deaths at the top = 3 columns
 - [x] I like the colorful map of COVID cases as the first figure below the topline numbers but I don't like the title.  It should say "COVID-19 Cases Map" and it should have the same font and alignment as the other titles (e.g., as in the "COVID-19 Cases Chart" title).
 - [x] Remove Indicators Summary line
 - [x] Remove Key Indicators

sgratzl:
closes #1042, closes #1054, closes #1052, closes #1048, closes #1043

**Prerequisites**:

- [x] Build is successful

### Summary

add hospital to the top list:

image

update list of signals
image

TODO:
 * some description texts are missing: https://docs.google.com/document/d/1llv6xh8jMlmVv7WpyDSv4VgUpFAZQ6QjFUeRxxOinmk/edit#
 * required https://github.com/cmu-delphi/delphi-epidata/issues/727


krivard:
> some description texts are missing: https://docs.google.com/document/d/1llv6xh8jMlmVv7WpyDSv4VgUpFAZQ6QjFUeRxxOinmk/edit#

COVID Act Now is in flux for the next two weeks -- they're changing their data source and removing support for total test volume, which will mess up our indicator. We should not add CAN until after the dust has settled.

sgratzl:
removed the PCR test from covid act now. thx for the doc update

krivard:
This seems a little bit silly (COVID/COVID, Quidel/Quidel)

image

I've dropped the duplicate Quidels from the signal name in the doc but I'm not sure what to do about the extra COVID. We need it in the source name to distinguish this group of signals from the Quidel flu tests, which aren't active... but I also don't want to remove it from the signal name, since we want to show other kinds of tests soon. Maybe just leave it in both places? @ryantibs thoughts?

ryantibs:
Just thinking about what's intuitive for a user: I think we should remove (COVID) at the end.  Since this is just labeling on our part, it doesn't need to exactly correspond in a 1:1 map sense with the source name.

That is, even if we were using "(Quidel Inc.)" as a post-fix for two signals that actually came from two different sources, I wouldn't be concerned, provided they're both Quidel of course, since it's just labeling on our part and we can do what seems most reasonable to the user.  

Even more, the flu source isn't active as Katie says, so I don't see any concern removing it.

krivard:
@ryantibs Apologies, I wasn't clear: One of the previous PRs switched the website from using a custom setting to describe the source for each signal, to using the centralized list we have in the advanced metadata endpoint at `epidata/covidcast/meta`. This centralized list reduces the effort required to add, remove, and maintain covidcast signals. **If we want to use a name for a source other than the one that's in the centralized list, it will make the website harder to maintain.**

I must not have noticed the reduplication in how the name of the quidel signal was displayed at the time, because [we've already merged it into the dev branch that way](https://dev--cmu-delphi-covidcast.netlify.app/?date=20211106) in preparation for the v3 release. Technically that means this change is out of scope for this PR, but depending on what we decide, we may be able to sneak it in.

The name of the source, currently "Quidel Inc. (COVID)" is one of two source subdivisions we make available from Quidel; the other is "Quidel Inc. (Flu)". The two subdivisions are separated because they have different geo and temporal availability patterns and active status. **Source subdivision names are required to be unique.** You can review the current list of sources and their subdivisions in the [meta sheet](https://docs.google.com/spreadsheets/d/1zb7ItJzY5oq1n-2xtvnPBiJu2L3AqmCKubrLkKJZVHs/edit#gid=852629429).

So the available options here are:
* Remove "(COVID)" from the source name by switching back to specifying source names for the website in a separate list, and just deal with having to keep multiple lists of sources up to date
* Remove "(COVID)" from the source name by merging the two Quidel subdivisions, and lose the ability to easily summarize availability and active status separately for Quidel COVID and Quidel Flu signals
* Live with "COVID Antigen Test Positivity (Quidel Inc. (COVID))"
* Switch to "Antigen Test Positivity (Quidel Inc. (COVID))"
* Something else

ryantibs:
Ok, thanks for the extra information.  I guess I was not aware of this general system (or else I forgot it if you told me about it before).

Given where we stand I think the simplest solution among the ones you propose is bullet point 3: Live with "COVID Antigen Test Positivity (Quidel Inc. (COVID))".  If we need to resolve this ASAP, then I'd say let's just go for that.

However, let me propose one more solution to consider.  If it's easy enough to implement and noncontroversial then we can try to sneak it in.  Otherwise please create a separate issue for us to consider it.

Alternative solution
===

Currently we print:

`signal_name` (`source_subdivision_name`)

Right?  So when the source subdivision name itself has a parentheses, it gets weird with nested parentheses.  This is what's happening here and it's also what would happen, it appears, with `chng-cli` and `smoothed_outpatient_cli`, which would print as:

COVID-Related Doctor Visits (Change Healthcare (COVID Like Illness))

which again seems a bit silly and repetitive.  So, in printing these things together: signal and source subdivision, why we don't make a rule that says that we just exclude anything that's in parentheses in the source subdivision name?  And if no parentheses exists then there is nothing to exclude.  So we'd get:

COVID Antigen Test Positivity (Quidel Inc.)
COVID-Related Doctor Visits (Change Healthcare)
COVID-Related Doctor Visits (Doctor Visits From Claims)
etc.

And in general, I think the way we're defining signal names and source subdivisions will make these combos unique and informative.  

And in other places, the source subdivision can be printed as-is, for other purposes (say to report on activity status, or whatever else they're used for).

ryantibs:
@sgratzl Can you take a look at my "Alternative solution" above and tell me how hard it is to implement?  If it's not hard I'd like to go with it.  If not then we should go with bullet point 3 in Katie's list.  Thank you.

sgratzl:
> @sgratzl Can you take a look at my "Alternative solution" above and tell me how hard it is to implement? If it's not hard I'd like to go with it. If not then we should go with bullet point 3 in Katie's list. Thank you.

would look like this:

image

ryantibs:
What you're showing looks like the alternative solution?  Looks good thanks.

ryantibs (commented):
It looks good to me but I noticed two things: 

1. I can't go past Nov 21 in the date selection at the top of the page (can't go to Nov 22, which I should be able to).  Maybe this is just a restriction that will be handled by the other PR. 
2. COVID-Like Symptoms (Delphi US COVID-19 Trends and Impact Survey) should *not* be under Public Behavior.  It should be under Early Indicators.

sgratzl:
> 1. I can't go past Nov 21 in the date selection at the top of the page (can't go to Nov 22, which I should be able to).  Maybe this is just a restriction that will be handled by the other PR.

is not related to the changes created by this PR, please report it separetly

> 2\. COVID-Like Symptoms (Delphi US COVID-19 Trends and Impact Survey) should _not_ be under Public Behavior.  It should be under Early Indicators.

that is part of the delphi-epidata meta data that is returned, see also https://docs.google.com/spreadsheets/d/1zb7ItJzY5oq1n-2xtvnPBiJu2L3AqmCKubrLkKJZVHs/edit#gid=852629429

ryantibs:
On 1, I reported it in a new issue.

On 2, ok thanks, I changed the google spreadsheet (I appeared to have permissions to do so). It still appears the same on the preview though. If that's too be expect, then I can approve this review so you can merge.

sgratzl:
> On 2, ok thanks, I changed the google spreadsheet (I appeared to have permissions to do so). It still appears the same on the preview though. If that's too be expect, then I can approve this PR so you can merge.

the spreadsheet is just an input for the delphi-epidata server. However, the file won't be used automatically. This PR https://github.com/cmu-delphi/delphi-epidata/pull/769/files has to be merged and then a new version of delphi-epidata released in order to be updated. So, it is the expected behavior what it is doing.

ryantibs:
Ok thanks. I will approve now (though to be clear, I want this categorization fix to occur with the v3 release)

ryantibs (approved):

PR #1069, opened 2021-11-08
fix showing rest of state tooltip

### LINKED ISSUE #1030: Rest-of-state hover tip has disappeared
(I wasn't sure whether to label this as a bug or a feature-request, but I was guessing it wasn't intentional, since we had this working previously and that seemed preferable because it gives the user more information about what they're seeing.)

In the choropleth maps on the "Explore and Indicator" site, the hover tip doesn't tell you when you're looking at a rest-of-state estimate.  The first screenshot shows the hover tip when the mouse is over a particular county, and the second shows the hover tip when the mouse is over MT's the rest-of-state.  

I will add that the fact that the parent state gets outlined when I hover over a county (as in MT gets outlined when I hover over Cascade County) is quite distracting to me, especially because this gray outline disappears when I move to a megacounty (rest-of-state).  I assume this will be fixed when we allow for the rest-of-state hover tip, since the gray outline will persist as I move around MT, instead of appearing and disappearing.

Screen Shot 2021-09-27 at 4 00 41 PM
Screen Shot 2021-09-27 at 4 08 04 PM

**Rating Scale (1 is minor, 7 is severe):** 2

sgratzl:
closes #1030 based on #1068 (for gdoc updates)

**Prerequisites**:

- [x] Build is successful

### Summary

fixes the missing rest of state tooltip:

image


krivard (approved):
👍

PR #1086, opened 2021-12-09
fix State Map for Hospital Admissions

### LINKED ISSUE #1083: State-level graphics for hospitalizations
I'd like us to properly support the state resolution for the HHS hospitalizations signal

1. Currently we don't show anything for the choropleth map on the "Explore an Indicator" page
Screen Shot 2021-12-08 at 11 32 39 AM

2. Currently the choropleth map on the "Explore a Location" page is also grayed out ....

Screen Shot 2021-12-08 at 11 37 46 AM

I'd like us just to default to showing the state-level choropleth map in both 1 & 2.  

(The color scale is also pretty bland, everything looks like it's the same color, but I think that is due to a bug in the way we're computing the hospitalizations signal, which will be my next issue.)

sgratzl:
closes #1083 

**Prerequisites**:

- [x] Build is successful

### Summary

image

show the state map when there are no counties data

### DIFF
--- a/src/blocks/RegionHexMap.svelte
+++ b/src/blocks/RegionHexMap.svelte
@@ -85,7 +85,7 @@
     interactive: false,
     domain: date.sparkLineTimeFrame.domain,
   });
-  $: colorScale = sensor.createColorScale(region.level);
+  $: colorScale = sensor.createColorScale('state');
   $: dumpData = Promise.all(tileData.map((d) => d.dump)).then((rows) => rows.flat());
 
   const maxColumn = state2TileCell.reduce((acc, v) => Math.max(acc, v.x), 0) + 1;
--- a/src/blocks/RegionMap.svelte
+++ b/src/blocks/RegionMap.svelte
@@ -30,15 +30,16 @@
    * @param {import("../../stores/params").RegionParam} region
    */
   function genSpec(sensor, region) {
+    const hasCounty = sensor.value.levels.includes('county');
     const options = {
-      domain: sensor.domain(region.level === 'state' || region.level === 'county' ? 'county' : 'state'),
+      domain: sensor.domain((region.level === 'state' || region.level === 'county') && hasCounty ? 'county' : 'state'),
       withStates: true,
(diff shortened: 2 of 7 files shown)


sgratzl:
re meta data: https://github.com/cmu-delphi/covidcast-indicators/issues/1399

ryantibs:
This looks good so far but I wanted to wait until the hospitalization signal is fixed (and metadata fixed accordingly) in #1399 so that we can check how the choropleth map looks like with the proper min/max before approving.

ryantibs:
I would say this is almost all fixed. The state map looks good as does the scale of the hospitalizations. Still a few problems remain (not related to the title of this PR but close enough---to do with the hospitalization signal):

1. The banner at the top says that the signal is not available for Dec 9, but it clearly is (numbers are shown directly below the banner): 

Screen Shot 2021-12-11 at 3 48 10 PM

2. On Dec 10, no hospitalization data is available so we get an NA, but the relative change says 0 (should be NA):

Screen Shot 2021-12-11 at 3 47 40 PM

sgratzl:
> I would say this is almost all fixed. The state map looks good as does the scale of the hospitalizations. Still a few problems remain (not related to the title of this PR but close enough---to do with the hospitalization signal):

should be fixed, please take a look

ryantibs:
The NA problem is fixed. The problem of the banner saying something isn't available when it actually is still persists (now shows up with cases):

Screen Shot 2021-12-11 at 8 28 38 PM

I seem to recall we had a previous issue and discussion on this (and blamed the metadata for being out of sync) but had implemented a general fix for it. Is this fix perhaps not just merged into the version I'm previewing?

ryantibs:
Thanks Sam---so, this seems to be a problem then somewhere in the backend and we should close this PR. Right? I will go ahead in approve it. 

But drawing attention to @krivard @korlaxxalrok @jgreene1959 because it sounds like they should know about the issues you raised.

ryantibs (approved):

PR #1096, opened 2021-12-09
Tune navigation hints

### LINKED ISSUE #1091: Improve visibility of subpage menu
Currently the only way to get the other dashboard subpages: "Explore an Indicator", "Indicator Status Overview", etc, is to click on the small box in the top right.  This is what is looks like on the dev site for me: 

Screen Shot 2021-12-08 at 8 53 56 PM

Would it look different when it gets launched?  I want to make sure we have a way of drawing visibility to this submenu.  Otherwise a person might think that "Explore a Location" (which is the landing page for the dashboard) may be the only page they get.

sgratzl:
closes #1091 

**Prerequisites**:

- [x] Build is successful

### Summary

image

image

image

what about this for better highlighting navigation options? @ryantibs

### DIFF
--- a/src/blocks/RegionMapWrapper.svelte
+++ b/src/blocks/RegionMapWrapper.svelte
@@ -1,4 +1,5 @@
 <script>
+  import mousePointerIcon from '!raw-loader!@fortawesome/fontawesome-free/svgs/regular/hand-pointer.svg';
   import RegionCountyMap from './RegionCountyMap.svelte';
   import RegionHexMap from './RegionHexMap.svelte';
   import RegionMap from './RegionMap.svelte';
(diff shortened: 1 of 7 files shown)


ryantibs:
Some aspects of this are nice but I find it a bit confusing because I don't think there's any way to get back to "Explore a Region" mode once you're in "Explore an Indicator" mode.  If I'm looking at a particular indicator and I click on a region then it takes me to the explore-an-indicator page for that region, not the explore-a-region page for that region.

So I think I would prefer to see some kind of explicit menu to allow the user to toggle between "Explore a Region" and "Explore an Indicator".

sgratzl:
ok, I added the navigation hints you mentioned in the issue:

image

ryantibs:
@sgratzl Looks good.  I'd like to replace "show details" with "explore further" (at the end of the sentences "Click on and region name..." and "Click on an indicator name...").  But then I'm ready to approve.

sgratzl:
done:
image

ryantibs:
Looking good!  

But I still see one more above the beehive/choropleth map: "Click on a state to show this region" --> should be "Click on a state to explore further".

ryantibs (approved):

PR #1114, opened 2022-01-05
feat: remove include zero toggle

### LINKED ISSUE #1112: Remove "Include 0 in Y Axis" toggle
I'd like us to consider removing the "Include 0 in Y Axis" toggle completely, on both the COVIDcast dashboard and the CTIS dashboard. 

I've never personally used it, and I've been using the dashboard a lot over the last 3 weeks. 

My concern is that it's just cluttering the toggle buttons that are already there (especially on mobile), and IMO it's the least important of them. 

Of course I'm only one type of user. And I know there was concern that people would be misinterpret things as trends when there is only tiny variation, without looking at the y-axis. But would simply having the option to include 0 in the y-axis really give them the hint that they should be cautious of this? 

All that taken together---and the fact that we don't really have a particular user we're targeting for the dashboards anyway---I don't think it's worth keeping it. 

Tagging @sgratzl @krivard @capnrefsmmat @RoniRos who are bound to have opinions about this 🙂

sgratzl:
closes #1112

**Prerequisites**:

- [x] Build is successful

### Summary
image

### DIFF
--- a/src/blocks/HistoryLineChart.svelte
+++ b/src/blocks/HistoryLineChart.svelte
@@ -95,14 +95,14 @@
    * @param {import('..stores/params').RegionParam} region
    * @param {import('../stores/params').DateParam} date
    * @param {import('../stores/params').TimeFrame} timeFrame
-   * @param {{height: number, zero: boolean, raw: boolean, isMobile: boolean, singleRegionOnly: boolean, stderr: boolean}} options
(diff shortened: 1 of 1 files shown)


ryantibs:
Thanks, looks good to me. I will wait to approve until we give a chance for a bit of discussion on #1112.

ryantibs (approved):
I think this has the desired behavior I was mentioning in my last comment on #1112, so, this is good to go.

PR #1119, opened 2022-01-06
remove short unit from google signal

### LINKED ISSUE #1117: Short postfix Google symptom search signals
It looks to me like we have "short" and "long" postfixes for our signals. 

Example of a long one: "per 100 visits", as in first screenshot. Example of short one: "per 100", as in second screenshot.

Screen Shot 2022-01-05 at 8 01 19 AM
Screen Shot 2022-01-05 at 8 01 56 AM

For the Google symptoms signal it looks like they're both: "scaled search volume". See two screenshots.

Screen Shot 2022-01-05 at 8 01 36 AM
Screen Shot 2022-01-05 at 8 02 07 AM

I'd like to change the short one, that shows up on as a hover tip, to be the empty string "". That is, all that would show up would be the number.

The rationale is that the short postfix in all the other cases is actually *not* displaying the units, it's just the normalization: "per 100" or "per 100k". And here, there is no normalization to mention.

Tagging @capnrefsmmat who will probably have an opinion about this.

sgratzl:
closes #1117 

**Prerequisites**:

- [x] Build is successful

### Summary

image

### DIFF
--- a/scripts/generateDescriptions.js
+++ b/scripts/generateDescriptions.js
@@ -113,6 +113,7 @@ function convertDescriptions(code) {
       yAxis: parseNestedOrString,
       casesOrDeathSignals: parseNestedOrString,
       mapTitleText: parseNestedOrString,
+      unitShort: (v) => v || '',
     });
   });
   fs.writeFileSync('./src/stores/descriptions.generated.json', JSON.stringify(entries, null, 2));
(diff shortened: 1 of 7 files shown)


ryantibs (approved):
Looks good, thank you.

PR #1132, opened 2022-01-13
extended scale for covid cases plots

### LINKED ISSUE #1131: Extend limits of color scale
(This should be the last item under v3.1!)

I think we're going to need to extend the limits of the color scale for the beehive & choropleth maps. Looking at cases right now, everything is dark red for the beehive map. Choropleth map is getting saturated in most of the south and southeast as well.

Screen Shot 2022-01-12 at 8 58 05 AM
Screen Shot 2022-01-12 at 8 58 16 AM
 
Can we just add a color to the end of the scale to keep the visualization consistent with what we have for everything below the max value on the color scale? That way the new color would only help distinguish the values about the max. For example, dark purple.

Also while I'm here, can I ask for a reminder of how these scales are computed. Is it manual or data-driven? It appears to be different for state resolution vs county resolution, for the same indicator. But just to double check: it is constant across all views over time and locations, right? Say, as I scroll back arbitrarily far into the past, or zoom into a particular location, the color scale would remain the same?

sgratzl:
closes #1131 

**Prerequisites**:

- [x] Build is successful

### Summary

see issue

please try it out, Currently it will only be done for covid-cases signal not for others.

### DIFF
--- a/src/blocks/RegionCountyMap.svelte
+++ b/src/blocks/RegionCountyMap.svelte
@@ -26,8 +26,7 @@
   export let fetcher;
 
   $: spec = generateStateMapWithCountyDataSpec({
-    domain: sensor.domain('county'),
-    scheme: sensor.value.vegaColorScale,
+    ...sensor.vegaSchemeDomain('county'),
   });
   $: data = fetcher.fetch1SensorNRegions1Date(sensor, 'county', date);
 
--- a/src/blocks/RegionMap.svelte
+++ b/src/blocks/RegionMap.svelte
@@ -32,9 +32,10 @@
   function genSpec(sensor, region) {
     const hasCounty = sensor.value.levels.includes('county');
     const options = {
-      domain: sensor.domain((region.level === 'state' || region.level === 'county') && hasCounty ? 'county' : 'state'),
+      ...sensor.vegaSchemeDomain(
+        (region.level === 'state' || region.level === 'county') && hasCounty ? 'county' : 'state',
+      ),
       withStates: true,
-      scheme: sensor.value.vegaColorScale,
     };
     if (region.level === 'state' && hasCounty) {
       return generateCountiesOfStateSpec(region.value, options);
--- a/src/components/HexGrid/ColorLegend.svelte
+++ b/src/components/HexGrid/ColorLegend.svelte
@@ -14,17 +14,19 @@
 
   $: scale = sensor.createColorScale(level);
 
-  $: linearScale = scale.copy().rangeRound([0, 100]);
+  $: linearScale = sensor.createValueScale(level).copy().rangeRound([0, 100]);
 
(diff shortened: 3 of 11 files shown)


ryantibs:
Thank you Sam!  To me this looks like a nice improvement and I like it.

I wonder whether I'm biased by the fact I've already familiarized myself with the color scale, having looked at many of these plots at this point over on the issue #1131. 

Before merging we may want get at least one more unbiased set of eyes to weigh in.  

@capnrefsmmat @krivard @RoniRos Can one of you quick look and see what you think of the extended color scale for cases?  [Click here for convenience](https://deploy-preview-1132--cmu-delphi-covidcast.netlify.app).

For example, is it confusing that light purple is "worse" than dark red?

When we get approval from at least one of these folks, then we can merge this and we should be ready to launch v3.1 !!

capnrefsmmat:
> For example, is it confusing that light purple is "worse" than dark red?

I think so, yes. I like the NYTimes scale because it's perceptually uniform, getting darker as it gets worse:

Screen Shot 2022-01-13 at 2 38 30 PM

When it goes past red, it goes into a dark reddish purple. In the preview for this PR, the purple is more of a nice lavender that's more blue than red, and is much lighter than the lower case values.

I think if you took this scale, removed the lavender part of it, and made the darker purples a bit more red, it'd look good.

RoniRos:
+1 Alex & Ryan (light purple > dark red is a problem.  NYT-like scale is good.)

ryantibs:
Thanks Alex. @sgratzl Can you try what Alex suggests?

krivard:
We should also be careful about contrast -- for example, in this screenshot OH, TX, and AK won't be legible for some users:

image

ryantibs:
I can try to pick specific colors at specific points, but can we try something easier first?

Can we define the extended color scale by just two colors its two ends, medium purple and dark purple.

(Whereas now you start at light purple.)

For medium purple you could use something similar to 2220 on the color scale in Katie's screenshot above, i.e., around the color of Kansas (KS).

For dark purple, the current end of the color scale, i.e., the color of Florida (FL). 

We can see how that looks----there might not be enough diversity in the gradient between those two end colors, but we can try.

sgratzl:
with

#7c187d to #4d014a

image
image

krivard (approved):
That looks amazing 🔥

ryantibs:
I think it looks good, it has the disadvantage that there's not as much diversity in the "extended range" so it's hard to tell those colors apart, but I think this is about as good as we can get right now, and we should move on.

Thanks for all the iteration Sam.

ryantibs (approved):

ryantibs:
@sgratzl @krivard This wraps up v3.1. So once this is merged, we should make a release.

PR #1136, opened 2022-01-18
multi region/indicator time series

### LINKED ISSUE #1113: Enable signal comparisons on time series plots
I'm using the dashboard frequently these days and I'm enjoying all the new fixes & simplifications. One thing that's missing for me is the ability to compare two signals at the same time, overlaid on a single time series plot. See the screenshot below as an example to build from:

Screen Shot 2022-01-04 at 5 11 38 PM

I would like to have a toggle called "Compare Indicators". This could show up, say, on the right-hand side of the plot. If this were toggled on, then it would allow me to select from any of the indicators (that are available at the given location) to put on the same time series plot. The default selection would be nothing.

- If a single other indicator is selected, then we would get dual y-axes, like the first screenshot below.
- If multiple other indicators are selected, then we would get signal values scaled to have the same min and max over the observed range, like the second screenshot below. 
- Moving your cursor over the time series would show all values for all signals at the given time value, like the third screenshot below.
- To be clear, there would be no other locations shown in faint lines anymore (in the current example, no more national, state, and average-of-neighbors curves). This would disappear as soon as we toggle on "Compare Indicators", which would only show the current location.

Screen Shot 2022-01-04 at 5 34 07 PM
Screen Shot 2022-01-04 at 5 40 07 PM
Screen Shot 2022-01-04 at 5 43 43 PM

As a separate idea, we could have another toggle called "Compare Locations". This would show up, say, on the right-hand side o fetch plot again, right above or below "Compare Indicators". If this were toggled on, then it would allow me to select from any of the locations (that are available for the given indicators). The default selection would be whatever the default is for the faint lines; that is, national, parent state, and average-of-neighbors for the current example.

- I think the y-axis business here would be very simple, we just keep it as is. 
- But the lines here are no longer faint, they are all painted fully/normally.
- And the hover tips would be as I suggested above in the "Compare Indicators" mode (shows values for all locations).

sgratzl:
closes #1113

**Prerequisites**:

- [x] Build is successful

### Summary

two new widgets for the dashboard builder: multi region and multi indicator time series:
depending on the selection it will be a single or dual axis:

image

### DIFF
--- a/src/blocks/SensorsLineTooltip.svelte
+++ b/src/blocks/SensorsLineTooltip.svelte
@@ -1,0 +1,58 @@
+<script>
+  import { formatDateShortDayOfWeekAbbr } from '../formats';
+  import SensorValue from '../components/SensorValue.svelte';
+
+  export let hidden = false;
+  /**
+   * @type {import('../../data').EpiDataRow}
+   */
+  export let item;
+
+  /**
+   * @type {View}
+   */
+  export let view;
+
+  /**
+   * @type {import('../../stores/params').Sensor[]}
+   */
+  export let sensors;
+
+  export let prop = 'sensorName';
+
+  $: items = item && view ? view.data('values').filter((d) => d.time_value === item.time_value) : [];
+</script>
+
+<div aria-label="tooltip" class="tooltip" class:hidden>
+  <h5>{formatDateShortDayOfWeekAbbr(item.date_value)}</h5>
+  <table>
+    {#each items as i, index}
+      <tr>
+        <th>{i[prop]}</th>
+        <td>
+          <SensorValue sensor={sensors[index]} value={i.value} medium />
+        </td>
+      </tr>
+    {/each}
+  </table>
+</div>
+
+<style>
+  .hidden {
+    display: none;
+  }
+
+  h5 {
+    margin: 0;
+    padding: 0;
+  }
+
+  th {
+    text-align: left;
+    max-width: 15em;
+  }
+  td {
+    text-align: right;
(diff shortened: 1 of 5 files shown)


ryantibs:
Thanks for tagging me, I had somehow not seen this when it first arrived.

I think it's a great addition and will be very useful. My thoughts below.

Main comment: it seems like we are limiting ourselves to 2 curves on either the multi-indicator or multi-region views (I tried selecting more than 2 but only 2 show up). I would like to allow 3 or more in both multi-indicator and multi-region modes. Let me explain.

- For multi-region, we should just have a single y-axis (not dual y-axis) since all signals will be of the same units anyway. If they're not comparable across geographies then that's why we have proportion versions (rates instead of counts). So I think it's not a main use case to want to be able to have separate axes for the same indicator over multiple regions. And we could simplify to a single y-axis, and then you should be able to add as many regions as you want to the plot.

- For multi-indicator, if the selection is of size 2, then I think it's nice to allow a dual y-axis. But you should be allowed to select more than 2, and then we just get a single axis and everything gets put onto an arbitrary scale from (say) 0 to 100, where for each signal, we scale it so that the min value over the selected time range being is 0, and so that the max value over the time range is 100. Then when you hover, you could see the actual values in original units for each signal.

Minor comment (unrelated to his PR, but it bugs me especially in the context of multiple signals): I would like us to remove "Rescale Y-Axis" radio button in general, because I think it's seldomly used (it doesn't appear anywhere else on our dashboards) and in the multi-signal context I find it confusing. I would expect rescaling to mean something else here.

PR #1140, opened 2022-01-19
Quidel Raw Hint

### LINKED ISSUE #1137: Quidel raw data availability seems misleading
@jingjtang @krivard @RoniRos @sgratzl Check this out:

COVIDcast  COVID Antigen Test Positivity Indicator Details 2
COVIDcast  COVID Antigen Test Positivity Indicator Details 4

There are long stretches of time where we have no raw data and yet there are smooth signal values available.  

I first thought this was a bug but now I think it's an artifact of the fact that we use a threshold of a count of 50 for raw signal availability, and yes for smoothed signals, we do geo-pooling if the count is 25 or higher and borrow from the parent state.

Anyway the result is visually quite confusing/misleading.  Can we add a small disclaimer below each of the Quidel time series graphs when in "Show Raw Data" mode, saying something like (paraphrasing): The way we do smoothing via geo-aggregation can result in a gross mismatch between raw signal availability and smooth signal availability.

sgratzl:
closes #1137 

**Prerequisites**:

- [x] Build is successful

### Summary

preparation for a data anomaly over all time that is shown for the raw quidel signal. 

e.g. an entry like:
```
{
        problem: 'Test Test',
        explanation:
          'The way we do smoothing via geo-aggregation can result in a gross mismatch between raw signal availability and smooth signal availability.',
        source: 'quidel-covid-ag',
        signals: 'covid_ag_raw_pct_positive',
        dates: '',
        regions: 'state(*);county(*)',
      }
```

Note: the format should be compatible with existing format but better be save and add it once it is deployed.

will result in

image

### DIFF
--- a/src/blocks/HistoryLineChart.svelte
+++ b/src/blocks/HistoryLineChart.svelte
@@ -274,7 +274,15 @@
 
   $: raw = singleRaw && sensor.rawValue != null && !($isMobileDevice && showFull);
   $: regions = raw ? [region.value] : resolveRegions(region.value, singleRegionOnly, showNeighbors);
-  $: annotations = $annotationManager.getWindowAnnotations(sensor.value, regions, timeFrame.min, timeFrame.max, true);
+  $: annotations = raw
+    ? $annotationManager.getMultiWindowAnnotations(
+        [sensor.value, sensor.rawValue],
+        regions,
+        timeFrame.min,
+        timeFrame.max,
+        true,
+      )
+    : $annotationManager.getWindowAnnotations(sensor.value, regions, timeFrame.min, timeFrame.max, true);
   $: spec = injectRanges(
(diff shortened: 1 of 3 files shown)


sgratzl:
re no warning: this PR is only the preparation. As noted in the PR description is requires a data anomaly entry in the google docs sheets that can be displayed. 

re other behavior. Please extract to an own issue, since not related to this PR

krivard:
yup. @ryantibs I was hoping for your review of sam's screenshot in the description

ryantibs:
Ok, that wasn't clear to me (next time it would help to be very explicit what you're looking for, from me). 

From what I can tell, the screenshot looks perfectly fine (ignoring the precise text since we can tweak that in the spreadsheets).  Thanks.

ryantibs (approved):

krivard (approved):
👍

PR #1146, opened 2022-01-21
Derive max date/issue for picker and hint

### LINKED ISSUE #1143: Inconsistency and/or lack of clarity in date definition on COVIDcast dashboard
When pulling down the date chooser, it says "the most recent available date is X, available on Y".  

1. In the example below, this appears incorrect: X=Jan 13, Y=Jan 18, but data seems available thru Jan 14.  This seems to be the case for many examples.

2. Even when the date is correct, the above quoted sentence is not as clear as can be.  How about something like:
      "Data is available for dates up through X (last updated on Y)"

image

sgratzl:
closes #1143 

**Prerequisites**:

- [x] Build is successful

### Summary

another workaround for determining the real max date and max issue

image

### DIFF
--- a/src/blocks/MaxDateHint.svelte
+++ b/src/blocks/MaxDateHint.svelte
@@ -10,49 +10,27 @@
   export let suffix = '';
 
   /**
-   * @type {import("../../stores/params").DateParam | null}
-   */
-  export let date = null;
-  /**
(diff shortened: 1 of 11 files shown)


krivard (approved):
nice!

PR #1165, opened 2022-03-03
Age Groups

### LINKED ISSUE #1138: Age support for age stratification
Our newest Quidel test positivity signals will be broken down by age groups: 0-4, 5-17, 0-17, 18-49, 50-64, 65+.  That is, these will be provided in addition to the original all-ages signal.

I would like the time series plots for Quidel test positivity on the COVIDcast dashboard to allow for viewing the signal broken down by age group.  I'm thinking something simple like this: 

COVIDcast  COVID Antigen Test Positivity Indicator Details

- We add a side toggle called "Age Groups", which defaults to being "off", so the default view is as it is now.  
- There are checkboxes below it that are all disabled (grayed out) and unchecked when the "Age Groups" toggle is off.  
- There is one checkbox for the "All Ages" group and then one checkbox per proper age group.  Each one also shows the color of the corresponding line.  (Colors here are just chosen as a demo; I'm not suggesting we actually use them.)
- When "Age Groups" is toggled "on", then you all these checkboxes are enabled and checked by default.  So you get the "All Ages" signal and every proper age group signal all overlaid, on the time series plot.
- When "Age Groups" is toggled "on", then the "Raw Data" toggle beneath the plot is disabled (no way to display raw data and multiple signals at once, would be too crowded), but the "Show All Dates" toggle is still enabled.  
- Note: while we're at it, I'd like to rename "Show All Dates" to be "All Dates", throughout the dashboard.  That would be consistent with "Raw Data" and "Age Groups" (where we don't start with a verb).  Also, to be clear, the "Include 0 in Y-Axis" toggle will be going away with v3.1, which hasn't been pushed live yet, and I got this screenshot from the live dashboard.

@sgratzl Do you need any other points of clarification and can you give it a shot?

sgratzl:
closes #1138 

**Prerequisites**:

- [x] Build is successful

### Summary

image

similar to the representation of how other regions (neighboring, state, ...) are shown.

### DIFF
--- a/scripts/generateDescriptions.js
+++ b/scripts/generateDescriptions.js
@@ -114,6 +114,7 @@ function convertDescriptions(code) {
       casesOrDeathSignals: parseNestedOrString,
       mapTitleText: parseNestedOrString,
       unitShort: (v) => v || '',
+      ageStratifications: parseArray,
     });
   });
   fs.writeFileSync('./src/stores/descriptions.generated.json', JSON.stringify(entries, null, 2));
--- a/src/blocks/HistoryLineChart.svelte
+++ b/src/blocks/HistoryLineChart.svelte
@@ -28,6 +28,7 @@
   import { TimeFrame } from '../stores/params';
   import { timeDay } from 'd3-time';
   import { resolveAgeStratifications } from '../stores/constants';
+  import SensorUnit from '../components/SensorUnit.svelte';
 
   export let height = 250;
 
@@ -128,7 +129,8 @@
       xTitle: sensor.xAxis,
       title: joinTitle([sensor.name, `in ${region.displayName}`], isMobile),
       subTitle: sensor.unit,
-      highlightRegion: showAge ? 'signal' : true,
+      highlightLine: !showAge,
+      filterLine: showAge ? 'signal' : false,
       isWeeklySignal: isWeekly,
       stderr,
     };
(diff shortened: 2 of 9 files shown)


ryantibs:
Thanks for getting started on this! Really cool to finally get to see these signals and play with them on the dashboard.

As per the design itself. Could you try implementing what I mocked up in #1138? The reason I  suggested that (and that I like that idea) is that it separates out the way we show multiple curves across locations (bottom legend) from multiple curves across age groups (right legend).

Also, the way it is now, there is just a lot of space taken up below the plot by the age group legend.

krivard:
I agree with Ryan that we want to visually distinguish between regional and age group stratification, but I also agree with Sam that this should display at the bottom so that we avoid screwing with the margin flow. I know mobile aren't our first priority users but I'd rather not destroy their experience if it's easily avoidable. We're also losing the orderedness by splitting into two columns.

We can stay in the same family of visual styles while highlighting the region-agegroup hierarchy and maintaining orderedness using something like

image

sgratzl:
what about: 

image

ryantibs:
I think it looks great!!

sgratzl:
> I think it looks great!!

is there then anything missing for this PR to be approved?

ryantibs:
I didn't realize you had already implemented it --- I assumed that was just a graphic or screenshot of work in progress.

Thanks Sam. I approved it.

ryantibs (approved):

PR #1209, opened 2023-02-10
Use JIT branch

dshemetov:
**Prerequisites**:

- [ ] Build is successful

### Summary

Uses cmu-delphi/delphi-epidata#1073

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -5,7 +5,7 @@
   COVIDCAST_ENDPOINT_URL = "https://api.covidcast.cmu.edu/epidata"
 
 [context.deploy-preview.environment]
-  COVIDCAST_ENDPOINT_URL = "https://api.covidcast.cmu.edu/epidata"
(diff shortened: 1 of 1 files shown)


dshemetov:
Seems to be pretty performant with just a little time spent clicking around. Any particular views that tend to be taxing on the dashboard? @krivard @melange396

krivard:
"Show all dates" for cases/deaths/hospital admissions is the greedy one.

I clicked around a little but I'm unfamiliar with the base dataset -- what date ranges and geos do we expect to be available, and are gaps expected or should everything be contiguous?

image
image

dshemetov:
That's a good question. It all depends on the data backup that the QA server is using. 

This version of JIT will not produce contiguous smoothed lines if the original data is not contiguous, so I'm guessing there's missing data there. @korlaxxalrok do you remember how the QA db was loaded?

PR #1227, opened 2023-09-22
Remove inactive sources from indicator status dashboard

rzats:
**Prerequisites**:

- [x] Build is successful
- [X] Code is cleaned up and formatted

### Summary

Prevents the inactive/stale sensors from showing up in the indicator status dashboard. Before:

Screenshot 2023-09-22 at 18 11 38

After:

Screenshot 2023-09-22 at 18 11 04

### DIFF
--- a/src/modes/indicator-status/IndicatorStatusTable.svelte
+++ b/src/modes/indicator-status/IndicatorStatusTable.svelte
@@ -9,6 +9,7 @@
   import chevronRightIcon from '!raw-loader!@fortawesome/fontawesome-free/svgs/solid/chevron-right.svg';
   import { metaDataManager } from '../../stores';
   import { loadData } from './data';
+  import { INACTIVE_DATA_SOURCES } from './store';
 
   const dispatch = createEventDispatcher();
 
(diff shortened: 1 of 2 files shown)


melange396 (commented):
This "works" but its kind of a hacky / band-aid solution...  Can we get useful information from available resources here instead of falling back on a hardcoded list of sources to exclude?

rzats:
@melange396 that information is already used [here](https://github.com/cmu-delphi/www-covidcast/pull/1227/files#diff-486ff657139eb4720bdfd99c4d1c90c971cb2c278736bba9a16d076fc9bf3222R16), where we filter out a bunch of sensors where the `active` property in the metadata is set to `false`. However, for the inactive sensors currently on the dashboard, that property is `true`.

A "smarter" solution could be to dynamically filter out all sensors that are N days out of date, which is similar to the process used to create the hardcoded list in the first place. I'd suggest a 180-day threshold in that case.

melange396:
We got rid of the inactive sources by updating the signal metadata in https://github.com/cmu-delphi/delphi-epidata/pull/1259#issuecomment-1758071698 and https://github.com/cmu-delphi/delphi-epidata/pull/1306#issuecomment-1758124724.

The change in this PR would hide signals marked active that havent been updated for ~half a year, which is good for keeping the dashboard clean but it might obscure signals that we need to take action on (like marking them inactive or by fixing them in some other way).  I'm going to close this for now, but we can revisit it later if something similar comes up.

PR #1235, opened 2023-11-02
Dashboard signal swaps

melange396:
Removing now-defunct JHU signals from the dashboard: replacing the deaths signal with one from `nchs-mortality` (which hasnt been used in the dashboard before), and replacing the cases signal with `doctor-visits`.


dsweber2 (approved):
preview seems good!

Issue #1236, opened 2023-11-03
Re-enable summary/Overview.svelte widget on dashboard


melange396:
Reenable the "Overview" widget that was commented out in `src/modes/summary/Summary.svelte` as part of #1235.  When the national-level death signal is fixed in https://github.com/cmu-delphi/covidcast-indicators/issues/1906 , the widget should no longer display "N/A" on initial load.

Alternatively, just delete the Overview widget...  But note that it encapsulates other widgets useful for the dashboard, like [`IndicatorWarning`](https://github.com/cmu-delphi/www-covidcast/blob/10bac380bfec7ee7a33a9b9b9929ba96592eba45/src/modes/summary/Overview.svelte#L45) (defined [here](https://github.com/cmu-delphi/www-covidcast/blob/dev/src/blocks/IndicatorWarning.svelte)) which provides helpful information in the case that there is a data lapse as well as an actionable link to a time frame with valid data.

  - also related:
    - https://github.com/cmu-delphi/www-covidcast/issues/1233
    - https://github.com/cmu-delphi/www-covidcast/pull/1234

melange396:
this is taken care of in #1243 , and we can close this issue when that PR is merged.

PR #1242, opened 2024-02-13
reenabling summary overview widget

melange396:
for #1236


melange396:
this is taken care of by #1243 , but i am going to leave this PR open until #1243 is merged, so we can see the netlify preview for reference.

PR #1244, opened 2024-03-14
Set COVIDcast export values based on URL parameters

rzats:
**Prerequisites**:

- [x] Build is successful

### Summary

Fills in the form values on the COVIDcast Export Data page based on URL parameters. This will be used to integrate the export UI with e.g. the EpiVis dashboard or signal documentation, with certain values being pre-filled based on parameters selected on another section of the website.

To test this out, visit the export dashboard on the preview website, then append this set of parameters to the URL:
```
/?data_source=covid-act-now&signal=covid-act-now-pcr_specimen_total_tests&start_day=2024-01-01&end_day=2024-02-02&geo_type=state
```

### DIFF
--- a/src/modes/exportdata/ExportData.svelte
+++ b/src/modes/exportdata/ExportData.svelte
@@ -56,8 +56,8 @@
     endDate = urlParams.has('end_day') ? new Date(urlParams.get('end_day')) : param.sparkLineTimeFrame.max;
 
     // Also normalize the dates to the current timezone
-    startDate = new Date(startDate.getTime() + Math.abs(startDate.getTimezoneOffset() * 60000));
-    endDate = new Date(endDate.getTime() + Math.abs(endDate.getTimezoneOffset() * 60000));
+    startDate = new Date(startDate.getTime() - startDate.getTimezoneOffset() * -60000);
+    endDate = new Date(endDate.getTime() - endDate.getTimezoneOffset() * -60000);
   }
   $: initDate($currentDateObject);
 
@@ -122,7 +122,7 @@
     asOfMode = 'single';
     asOfDate = new Date(urlParams.get('as_of'));
     // Also normalize the dates to the current timezone
-    asOfDate = new Date(asOfDate.getTime() + Math.abs(asOfDate.getTimezoneOffset() * 60000));
+    asOfDate = new Date(asOfDate.getTime() - asOfDate.getTimezoneOffset() * -60000);
   }
 
   let form = null;
--- a/src/modes/exportdata/ExportData.svelte
+++ b/src/modes/exportdata/ExportData.svelte
@@ -54,6 +54,10 @@
     // Populate date based on URL params or, if absent, the current date
     startDate = urlParams.has('start_day') ? new Date(urlParams.get('start_day')) : param.sparkLineTimeFrame.min;
     endDate = urlParams.has('end_day') ? new Date(urlParams.get('end_day')) : param.sparkLineTimeFrame.max;
+
+    // Also normalize the dates to the current timezone
+    startDate = new Date(startDate.getTime() + Math.abs(startDate.getTimezoneOffset() * 60000));
+    endDate = new Date(endDate.getTime() + Math.abs(endDate.getTimezoneOffset() * 60000));
   }
   $: initDate($currentDateObject);
 
@@ -117,6 +121,8 @@
   if (urlParams.has('as_of') && !isNaN(new Date(urlParams.get('as_of')))) {
     asOfMode = 'single';
     asOfDate = new Date(urlParams.get('as_of'));
+    // Also normalize the dates to the current timezone
+    asOfDate = new Date(asOfDate.getTime() + Math.abs(asOfDate.getTimezoneOffset() * 60000));
   }
 
   let form = null;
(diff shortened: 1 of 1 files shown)


melange396 (changes requested):
Not bad so far!

 - Date field population does not seem to be working.
 - We should also allow specification of `as_of` as well, if thats easy enough to do (ie, if its done the same way as the other date fields).
 - Can we use the same url parameters as the API uses? (like `geo_value` instead of `geo_id`)
 - Add an optional field for an api key to allow access to restricted signals (in the ui as well as in the url).  The key should also be used to retrieve metadata.
 - The selected data source is not shown when specifying an "inactive" source (see https://deploy-preview-1244--cmu-delphi-covidcast.netlify.app/export/?source=covid-act-now&sensor=covid-act-now-pcr_specimen_total_tests&start=2024-01-01&end=2024-02-02&geo_type=state )

melange396:
clarification: the dates in the url and in the page seem to be off by one for me --
image

rzats:
Notes from the meeting:
- The initial date seems to be set fine but then the datepicker breaks - investigate that
- Names for URL parameters: https://cmu-delphi.github.io/delphi-epidata/api/covidcast.html
- API key URL field: put this before step 0 to get access to private signals

rzats:
@melange396 fixed up all your suggestions (except the API key to allow access to restricted signals - I think that's best done in a separate PR, since it's not really related to URL params)

Here's a new URL to test out the fix to the issue you've mentioned:
https://deploy-preview-1244--cmu-delphi-covidcast.netlify.app/export/?data_source=covid-act-now&signal=covid-act-now-pcr_specimen_total_tests&start_day=2024-01-01&end_day=2024-02-02&geo_type=state

In short - this was a bug with inactive sensors in particular, where the datetime selector was incorrectly set to current (e.g. 2024) dates for inactive sensors (so e.g. only running between 2020-2021). It was possible to reset the dates to valid ones by re-selecting the date on the picker, but a bit unintuitive.

I fixed this by setting the datetime selector to the sensor's minimum and maximum start time, if it's inactive.

src/modes/exportdata/ExportData.svelte
```diff
@@ -27,18 +27,55 @@
...
   function initDate(date) {
     const param = new DateParam(date);
-    endDate = param.sparkLineTimeFrame.max;
-    startDate = param.sparkLineTimeFrame.min;
+
+    // Populate date based on URL params or, if absent, the current date
+    if (sensor && !sensor.active) {
+      // If the sensor is inactive, set the start and end dates to its historical range rather than current date
+      if (
+        urlParams.has('start_day') &&
```
melange396, line 58: Im not sure we want to clamp these values to the active range if theyve been specified in the URL...  The principle of least surprise suggests showing the user what they asked for.

melange396 (dismissed):

src/modes/exportdata/ExportData.svelte
```diff
@@ -27,18 +27,37 @@
...
-    endDate = param.sparkLineTimeFrame.max;
-    startDate = param.sparkLineTimeFrame.min;
+
+    // Populate date based on URL params or, if absent, the current date
+    startDate = urlParams.has('start_day') ? new Date(urlParams.get('start_day')) : param.sparkLineTimeFrame.min;
+    endDate = urlParams.has('end_day') ? new Date(urlParams.get('end_day')) : param.sparkLineTimeFrame.max;
+
+    // Also normalize the dates to the current timezone
+    startDate = new Date(startDate.getTime() - startDate.getTimezoneOffset() * -60000);
+    endDate = new Date(endDate.getTime() - endDate.getTimezoneOffset() * -60000);
```
dshemetov, line 60: Let's add a comment about the Javascript string parsing + negative timezone edge cases here. If this is a common fix, maybe we can include a link instead of writing the full explanation out.

dshemetov (approved):
This looks good enough to merge to me. Since this feature will have a "silent" release for now and none of the changes are hard to revert or modify, we can keep iterating on it live.

After syncing with Rostyslav on Slack and trying out some features out in the browser, we have a list of a few outstanding issues that Rostyslav will address in the next PR:
* `geo_value=CA` is allowed, but `geo_value=ca` causes the page to hang; this is the opposite of the API behavior and needs a workaround (like dynamically lower-casing or upper-casing as needed)
* when browsing on the dashboard the main URL parameters I have accumulated were `sensor`, `region`, and `date`; we will need to convert `sensor` into `data_source` and `signal` (should be doable with a split on "-") and then make sure that `region` gets correctly converted to `geo_value` and `date` gets correctly converted to `start_date` / `end_date`

Let's make sure those tasks are captured in whatever issue is tracking this work.

melange396:
Why was this merged?  There are still unaddressed concerns, plus there is no reason to release bugged and undertested functionality.

dshemetov:
I thought Rostyslav addressed most of the concerns from your previous review? My personal preference here is to not let the perfect be the enemy of the good. Rostyslav made an [issue on Jira](https://cmu-delphi.atlassian.net/browse/OKRS24-161) to track TODOs on this feature.

rzats:
Things to fix up:
- `geoURLSet` - try bringing it out of the `$` block so it only runs once
- `geo_value` - convert it to uppercase internally
- Add code comment about timezone conversions, especially explain the weird double negative

dshemetov:
Also this one:
* when browsing on the dashboard, the URL parameters I noticed were `sensor`, `region`, and `date`; we will need to convert `sensor` into `data_source` and `signal` (should be doable with a split on "-"), `region` into `geo_value` and `date` into `start_date` / `end_date`

PR #1250, opened 2024-04-24
Update GH Actions versions to match those in www-main

melange396:
inspired by https://github.com/cmu-delphi/www-main/pull/959


rzats (approved):
lgtm!

PR #1271, opened 2025-03-20
update gh 'actions/cache' away from deprecated version

melange396:
The version currently referenced causes actions to fail:
https://github.com/cmu-delphi/www-covidcast/actions/runs/13954044807/job/39081797085#step:1:37

This PR is very similar/analogous to https://github.com/cmu-delphi/delphi-epidata/pull/1603 .


dmytrotsko (approved):
:+1:

